PR #3, opened 2022-05-03
Feature/feature table core units

alejandrocabriales:
# Ticket
[Trello-19 Layout CU Table](https://trello.com/c/V3zDi4b4/19-layout-cu-table)

# Description
Creating table for the Core Units Table on the Dashboard

# Actions
Created  Table for core Units
Created Item Header
Created Header
Create Icons Arrows
add Material Ui Core
Add table to storybook


package.json
```diff
@@ -6,6 +6,7 @@
     "@apollo/client": "^3.6.0",
     "@emotion/react": "^11.9.0",
     "@emotion/styled": "^11.8.1",
+    "@material-ui/core": "^4.12.4",
```
z3phyro, line 9: Seems like a previous version of material was installed by error here

z3phyro (commented):
If possible let's move all components to the stories in storybook.

src/components/Svg/ArrowDown.tsx
```diff
@@ -1,5 +1,5 @@
 import * as React from 'react';
-
+/* eslint-disable */
```
z3phyro, line 2: If there isn't any impediment let's remove this eslint comment

package.json
alejandrocabriales: you are right, i will update  to the last

PR #6, opened 2022-05-05
Trello 19 layout cu table

z3phyro:
# Ticket 
[Trello-19](https://trello.com/c/V3zDi4b4/19-layout-cu-table)

# Actions
Added Links column layout
Added FTE parameter to members column
Added Inter font to solution
Added budget cap to expenditures
Added popover


src/stories/components/custom-popover/custom-popover.scss
```diff
@@ -0,0 +1,4 @@
+.MuiPaper-root {
+  border-radius: 10px !important;
```
alejandrocabriales, line 2: I don't if is possible not use import, if is necessary is ok, but we should try to avoid to use

z3phyro: Solved

alejandrocabriales (approved):

PR #15, opened 2022-05-11
change-md-component-use-text-insted-url

alejandrocabriales:
# Ticket
[Trello-20-layout-cu-about](https://trello.com/c/JqoadMnP/20-layout-cu-about)

# Description
The "CU About" view will be presented once the user clicks on the "Core Unit -> About" option in 

 # Actions
 Change the md-view components to work with text instead of url


z3phyro (approved):
Looks good to me

PR #17, opened 2022-05-17
SES-19 layout cu table fixes

z3phyro:
# Ticket
[SES-19](https://trello.com/b/iebIfrww/ses-dashboard)

# Actions
Removed extra colon on title
Fixed top right icons size
Fixed repeated linked in 
Fixed filters height 
Fixed status chip font size  and height
Checked chip border radius and size


src/stories/components/svg/feedback.tsx
```diff
@@ -20,7 +20,7 @@ const FeedBack = ({
       onClick={onClick}
       width={width}
       height={height}
-      viewBox='0 0 23 23'
+      viewBox={`0 0 ${width} ${height}`}
       fill='none'
```
alejandrocabriales, line 24: We can pass also the fill, that way we can use the same icon , with different color. What do you think?

PR #33, opened 2022-05-27
[SES-47] Created new custom select

z3phyro:
# Ticket
[SES-47](https://trello.com/c/DzZsWfLA/47-refine-the-layout-of-cu-table)

# Actions
Created new custom select to match new design

### DIFF
--- a/src/stories/components/custom-multi-select-2/custom-multi-select-2.tsx
+++ b/src/stories/components/custom-multi-select-2/custom-multi-select-2.tsx
@@ -1,0 +1,124 @@
+import React, { useRef, useState } from 'react';
+import styled from '@emotion/styled';
+import { SelectChevronDown } from '../svg/select-chevron-down';
+import { Checkbox, ListItemText, MenuItem, Typography } from '@mui/material';
+import CheckBoxOutlined from '@mui/icons-material/CheckBoxOutlined';
+import './custom-multi-select-2.scss';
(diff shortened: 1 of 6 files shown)


src/stories/components/custom-multi-select-2/custom-multi-select-2.tsx
```diff
@@ -0,0 +1,124 @@
...
+      setActiveItems(temp);
+    }
+  };
+
+  const toggleVisible = () => setVisible(!visible);
+
+  const toggleAll = () => {
+    if (activeItems.length === props.items.length) {
+      setActiveItems([]);
+      props.onChange && props.onChange([]);
```
alejandrocabriales, line 45: I think you can destructuring the props that you are going to need, then you can avoid write props.something, of course its not necessary, but for me look much easy to read

alejandrocabriales:
Just a small comment to see your point, the rest look to me

src/stories/components/custom-multi-select-2/custom-multi-select-2.tsx
z3phyro: TBH I don't agree. Having the properties inside the props object helps with clarity, it's easier to identify which variables are local and which come from the father component. Personally I only use "spread" when assigning default values to it, but it was good to get your point of view from it.

PR #34, opened 2022-05-27
not-show-date-when-incorrect-data-backed

alejandrocabriales:
Not show data when some incorrect data came from backed


alejandrocabriales:
This is for not show data, when status in mip don't fit whit date

PR #36, opened 2022-05-31
SES unit test core unit about functions

alejandrocabriales:
# Ticket
Unit test some fuctionality

# Description

# Actions
#########SES-Unit-Test-Core-Unit-About-Functions#########
Add firts commit function getMipsStatus
Add function for get each markdown text
Add test for getLinkFrom contributor
Fix mips should be order before ask status
Delete duplicate-function and extends the other one


z3phyro (approved):
Looks good to me

PR #52, opened 2022-06-16
Ses 49 create fidelity design of cu about

alejandrocabriales:
# Description

# Actions
Create higth fidelity:
SES-49-create-fidelity-design-of-cu-about
Add scroll in the card navegation change color icons
reset-unnecesary-comment
Add scrolling bar for navegation bar and add story for Cu-about


z3phyro (approved):
LGTM

PR #56, opened 2022-06-21
Migrate to nextjs

alejandrocabriales:
This add router insted react-router-dom.
Also work with the page for next


z3phyro (approved):
LGTM

PR #93, opened 2022-07-06
add-wrapper-also-border-correct-div

alejandrocabriales:
This pr is for resolve issues  relate with header
Quick fix to problem find in the meeting with the client


z3phyro (approved):
LGTM

PR #98, opened 2022-07-11
Ses fix cu landing page final iteration

alejandrocabriales:
This add correct padding when the description have two lines


z3phyro (approved):
LGTM

PR #102, opened 2022-07-12
add-letter-correct-position-after-change-position-header

alejandrocabriales:
# Ticket
No ticket

# Description
Some fixed in the layout
# Actions
add-letter-correct-position-after-change-position-header


z3phyro (approved):
LGTM

PR #130, opened 2022-07-25
[SES-89] implement responsive solution for cu index table

z3phyro:
# Ticket
https://trello.com/c/GOwDLF4X/89-implement-a-responsive-solution-for-cu-index-table

# Description
Added multi-select exact design, responsiveness and functionality


src/stories/components/select-item/select-item.tsx
```diff
@@ -0,0 +1,61 @@
...
+  checked?: boolean;
+  onClick?: () => void;
+  minWidth?: number;
+}
+
+export const SelectItem = ({ checked = false, minWidth = 0, ...props }: SelectItemProps) => {
+  const [focused, setFocused] = useState(false);
+
+  return <Container className="no-select" onClick={props.onClick} minWidth={minWidth}>
+    {!checked && <CheckboxOff style={{ padding: '2px' }} fill={focused ? '#708390' : '#9FAFB9'}/>}
```
alejandrocabriales, line 18: Just and ide, what do you thing to use ternary operator in here, give me your feedback about what do you think

.eslintrc.js
```diff
@@ -22,5 +22,6 @@ module.exports = {
     'space-before-function-paren': ['error', 'never'],
     'comma-dangle': ['error', 'only-multiline'],
     'object-property-newline': ['error', { allowAllPropertiesOnSameLine: false }],
+    '@typescript-eslint/no-explicit-any': 'off',
```
alejandrocabriales, line 25: What do you think about , just desirable in (eslint) in the line that it request, and then in some time, try to find the correct type. Give your opinion about it

PR #135, opened 2022-08-01
fix-three-first-items-space

alejandrocabriales:
#Not Ticket 
Fix the first three elements of the
85-general-qc-round-v-10-r


z3phyro (approved):
LGTM

PR #166, opened 2022-08-11
[SES-95-96] Added dynamic names of core units about and transparency

z3phyro:


z3phyro:
Code is not up to date

PR #173, opened 2022-08-12
Ses 83 ensure server side rendering for cu about page v 10

alejandrocabriales:
# Ticket
https://trello.com/c/aG5hksM7/83-ensure-server-side-rendering-for-coreunitaboutpage-v-10

# Description
Server Side Rendering

# Actions
Add props and server side function
Add correct tags for elements
Fix hydration problem and fix lint


pages/core-unit/[code]/index.tsx
```diff
@@ -1,11 +1,25 @@
...
-    <CuAboutContainer />
+    <CuAboutContainer code={code} cuAbout={cuAbout as CuAbout} contributors={contributors} />
   );
 };
-
 export default CoreUnitAboutPage;
+export const getServerSideProps: GetServerSideProps = async(context: GetServerSidePropsContext) => {
+  const { query } = context;
+  const code = query.code as string;
+  const cuAbout = await fetchCoreUnitByCode(code);
```
Comment (YasielCabrera) [inline pages/core-unit/[code]/index.tsx:16]: A 404 error should be thrown when there are any CoreUnit with the given code. Otherwise an empty page will be rendered

See https://nextjs.org/blog/next-10#notfound-support

YasielCabrera (commented):
I think 404 response should be handled in getServerSideProps to prevent empty pages when the users write the URL manually

pages/core-unit/[code]/index.tsx
Comment (alejandrocabriales) [reply pages/core-unit/[code]/index.tsx:16]: you are right , i will add it

pages/core-unit/[code]/index.tsx
```diff
@@ -1,11 +1,34 @@
...
+  const contributorCommitment = cuAbout.contributorCommitment;
+
+  console.log('cuAbout', cuAbout);
```
Comment (YasielCabrera) [inline pages/core-unit/[code]/index.tsx:20]: You forgot to remove the `console.log`. Also, there is no need to check if the `contributorCommitment` length is 0 since if `cuAbout` is empty, `contributorCommitment` will be empty too

Comment (alejandrocabriales) [reply pages/core-unit/[code]/index.tsx]: you are right i will remove both of them

YasielCabrera (approved):

PR #197, opened 2022-08-17
[SES-108] add sitemap

YasielCabrera:
# Ticket
https://trello.com/c/trpIoYYf/108-implement-seo-elements-v-10

# Description
A sitemap of all the existing urls was implemented. The sitemap can be disabled using feature flags according to the environment

# Actions
- created a sitemap builder to generate the sitemap content
- added two new environment variables
- cached sitemap content on the client side (1h)
- updated readme.md

### DIFF
--- a/src/config/routes.ts
+++ b/src/config/routes.ts
@@ -1,4 +1,1 @@
 export const BASE_URL = process.env.NEXT_PUBLIC_BASE_URL ?? 'https://expenses.makerdao.network';
-export const HOME_ROUTE = '/';
-export const ABOUT_ROUTE = '/about/:code';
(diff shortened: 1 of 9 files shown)


src/config/routes.ts
```diff
@@ -1,3 +1,4 @@
+export const BASE_URL = process.env.NEXT_PUBLIC_BASE_URL ?? 'https://expenses.makerdao.network';
 export const HOME_ROUTE = '/';
```
alejandrocabriales, line 2: I know that its not relate with the pr, but i think that you can remove unnecessary router in the file

YasielCabrera: Sure, I will

alejandrocabriales (approved):

PR #208, opened 2022-08-21
[SES-107]

z3phyro:
# Ticket
https://trello.com/c/KfjSlKiE/107-general-issues-detected-in-the-uat-r


src/stories/components/custom-popover/custom-popover.tsx
```diff
@@ -28,6 +35,9 @@ export const CustomPopover = ({
   const [leaveTimeout, setLeaveTimeout] = React.useState<NodeJS.Timeout>();
 
   const handlePopoverOpen = (event: React.MouseEvent<HTMLElement>) => {
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    document.querySelector('body').onscroll = () => handlePopoverClose();
```
YasielCabrera, line 40: The `onscroll` event should be removed after unmounting the Popover component otherwise `handlePopoverClose` will continue to be executed event after the popover closes, also multiple events could be registered and executed. You can check this by adding a `console.log` or any other debug method to the `handlePopoverClose` function and scroll the browser. I'm not entirely sure, but this should be solved by adding:

```js
useEffect(() => {
  return () => document.querySelector('body').removeEventListener('onscroll', handlePopoverClose);
});

// also replace the line:
// document.querySelector('body').onscroll = () => handlePopoverClose();
// by
// document.querySelector('body').onscroll = handlePopoverClose;
```
Please, note that the code above has not been tested

alejandrocabriales: i will check it out

YasielCabrera (approved):
LGTM

PR #218, opened 2022-08-26
[SES-110] Reflect the filter and sorting in the inside pagination of cu about

YasielCabrera:
# Ticket
https://trello.com/c/QJFWbzL0/110-general-issues-detected-in-the-uat-coreunitaboutpage-r

# Description
When a user filters on the core unit index table and then clicks into the specific core unit. The filter is not reflected when a user toggles back to the previous, or ahead to the next core unit, so this behavior was changed to reflect the filters

# Actions
- The ordering logic was moved to redux
- The sorting CU function was moved out of the CU table so it can be reused
- The pagination now works by applying the filters and orderings in the CU table


alejandrocabriales (approved):
LGTM

PR #250, opened 2022-09-04
[SES] Make CU about pages fully SSR

YasielCabrera:
# Ticket
NO TICKET

# Description
Moved all queries previously done on the client side to the Server Side

# Actions
- Moved queries from core-unit-summary to SSR in CU about page
- Moved queries from core-unit-summary to SSR on CU transparency report page
- Removed `budgetStatements` from the `core-unit-summary` query to reduce the weight of the data fetched
- Apply SSR to the markdown elements in the CU about page


alejandrocabriales (approved):
LGTM

PR #263, opened 2022-09-08
Ses 131 improvement 1 remove the cu about from storybook pages

alejandrocabriales:
# Ticket
SES-131-improvement-1-remove-the-cu-about-from-storybook-pages

# Description

# Actions
remove cu-about from storybook and delete unnecessary query contributors

fix-syntactic-error


z3phyro (approved):
LGTM

PR #264, opened 2022-09-09
SES 133 improvements 6 add a plugin to eslint to correct spelling problems

alejandrocabriales:
# Ticket
133-improvements-6-add-a-plugin-to-eslint-to-correct-spelling-problems-v-20

# Description

# Actions
Add regex and and word to eslintrcjs
fix-problem-core-uni-summary
Add eslint-plugin-spellcheck


.eslintrc.js
```diff
@@ -11,16 +12,120 @@ module.exports = {
...
+          'Hight',
+          'Subheader',
+          'Identicon',
+          'ident',
+          'Unordered',
+          'juan',
+          'Julien',
+          'wouter',
+          'Kampmann',
+          'Wouter',
```
z3phyro, line 105: Do you think these are needed on the skip words dictionary?

.eslintrc.js
```diff
@@ -11,16 +12,120 @@ module.exports = {
...
+          'Tooltip',
+          'Team’s',
+          'Petru',
+          'Catana',
+          'Jevans',
```
z3phyro, line 110: Do you think these are needed on the skip words array?

.eslintrc.js
```diff
@@ -11,16 +12,120 @@ module.exports = {
...
+          'onchain',
+          'Whitepaper',
+          'Onboarding',
+          'gtag',
+          'mailto',
+          'Millis',
+          'Datetime',
+          'Sitemap',
+          'Mip', // Explicity word to talk about mips
+        ],
```
z3phyro, line 120: Can we move this array to a different file .json to make the config more readable?

alejandrocabriales: you are right, I will move to array

PR #268, opened 2022-09-14
Ses 136 improvement 4 allow to have the cu about related api logic in one place v 20

alejandrocabriales:
# Ticket
136-improvement-4-allow-to-have-the-cu-about-related-api-logic-in-one-place-v-20

# Description

# Actions
Remove the all data and add correct DTO
Move function logical folder and add correct DTO
Delete unnecessary DTO and add commitment enum
Add correct DTO index page
Add lint-stage to husky
Fix-testing
Fix-error-in-test-yarn-lint
Fix-storybook
Add DTO to slice
Correct error in mock storybook
Move-contributors-same-query
Fix-error-after-mix-master


z3phyro (approved):
LGTM

PR #272, opened 2022-09-16
[SES-134] Remove warnings

z3phyro:
# Ticket
https://trello.com/c/0XgPWjxG/134-improvements-8-remove-all-the-warning-from-the-console-v-20


src/stories/components/404/not-found-component-404.tsx
```diff
@@ -182,7 +182,7 @@ const TextUps = styled(Typography)<{ isLight?: boolean }>(({ isLight }) => ({
   },
 }));
 
-const TextDescription = styled(Typography)({
+const TextDescription = styled(Typography, { shouldForwardProp: (prop) => prop !== 'isLight' })({
```
alejandrocabriales, line 185: I am in doubt relate with one, why should use shouldForwadProps if we are not going to use is Light props

z3phyro: Solved

PR #355, opened 2022-10-17
Add default fonts for the body

alejandrocabriales:
#No ticket
Add default Inter for the body

# Description

# Actions
Add default fonts for the body


z3phyro (changes requested):
Do we really need to do this array operation?

alejandrocabriales:
> Do we really need to do this array operation?

You are right, fixed

z3phyro (approved):
LGTM

PR #377, opened 2022-10-28
SES 188 feature core unit finances transparency reporting v3 API

alejandrocabriales:
# Ticket
https://trello.com/c/s87DZLZr/188-feature-coreunitfinancestransparencyreporting-v3

# Description

# Actions
Function get comments and number comments

Correct change from the API in the view

### DIFF
--- a/src/stories/containers/transparency-report/transparency-comments/comment-item.tsx
+++ b/src/stories/containers/transparency-report/transparency-comments/comment-item.tsx
@@ -3,11 +3,11 @@ import { DateTime } from 'luxon';
 import Markdown from 'marked-react';
 import lightTheme from '../../../../../styles/theme/light';
 import { useThemeContext } from '../../../../core/context/ThemeContext';
-import { CuCommentDto } from '../../../../core/models/dto/comments.dto';
+import { CommentsBudgetStatementDto } from '../../../../core/models/dto/core-unit.dto';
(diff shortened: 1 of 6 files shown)


src/stories/containers/transparency-report/transparency-comments/comment-item.tsx
```diff
@@ -16,9 +16,9 @@ const CommentItem = ({ comment, code }: Props) => {
   return (
     <CommentItemContainer isLight={isLight}>
       <UTCDate isLight={isLight}>
-        {`Comment on ${DateTime.fromISO(comment.commentDate ?? '')
+        {`Comment on ${DateTime.fromISO(comment.timestamp ?? '')
```
YasielCabrera, line 19: I can see that `timestamp` is a required field on `CommentsBudgetStatementDto` and you are expecting an empty value here, if this value can be empty the field in the DTO should be set as optional to avoid possible errors in other places where it can be used. Otherwise, the null check here is not needed. Do we have that info from the models in the API?

alejandrocabriales: The edge case could be that there's some internal server error...
In this case, no comment will be created
So provably right now is not necessary if the API work as comment about, but just to prevent

YasielCabrera (approved):

PR #379, opened 2022-10-31
[SES-158] qc round 3

z3phyro:
# Ticket
https://trello.com/c/VV4dX1cU/158-qc-round-v-30-r


src/stories/components/custom-multi-select/custom-multi-select.tsx
```diff
@@ -125,7 +125,7 @@ export const CustomMultiSelect = ({
               placeholder="Search"
               value={searchText}
               onChange={(val) => setSearchText(val)}
-              handleCleanSearch={() => setSearchText('')}
+              handleClearSearch={() => setSearchText('')}
```
alejandrocabriales, line 128: What do you think about avoid inline function:
you can use handleClearSearch, callback and call that function
const  handleClearSearch=useCallback(()=>{
setSearchText('')
})
What do you think ????

z3phyro: Let's do it

PR #382, opened 2022-11-01
Fix qc round 5

alejandrocabriales:
# Ticket
SES-159-qa-bug-findings-v-30-r

# Actions
Change name button place holder and add text button
Add correct gap and button text in mobile

### DIFF
--- a/src/stories/components/cu-activity-table/cu-activity-table.placeholder.tsx
+++ b/src/stories/components/cu-activity-table/cu-activity-table.placeholder.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useCallback } from 'react';
 import Image from 'next/image';
 import styled from '@emotion/styled';
 import PlaceholderImg from '../../../../public/assets/img/rectangles.png';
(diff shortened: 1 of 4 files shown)


src/stories/components/cu-activity-table/cu-activity-table.placeholder.tsx
```diff
@@ -37,9 +39,12 @@ export const ActivityPlaceholder = (props: { hasFilter: boolean }) => {
             buttonType={ButtonType.Secondary}
           />
         )}
-        <a href={SUBMIT_EXPENSES_URL} target="_blank">
-          <CustomButton style={{ minWidth: '250px' }} label="Submit Expenses Now" buttonType={ButtonType.Primary} />
-        </a>
+        <CustomButton
```
YasielCabrera, line 42: can this be a `Link` (from nextjs) to navigate instead a button with a `onClick` event?

src/stories/components/reset-button/reset-button.tsx
```diff
@@ -6,27 +6,44 @@ import { Close } from '../svg/close';
 interface Props {
   onClick?: () => void;
   disabled?: boolean;
+  label?: string;
+  isWithIcon?: boolean;
+  labelMobile?: string;
 }
 
-export default (props: Props) => {
+export default ({ onClick, disabled, label = 'Reset Filters', isWithIcon = true, labelMobile }: Props) => {
```
YasielCabrera, line 14: we can use `hasIcon` here as it is less verbose and still refers to a boolean according to the conventions

YasielCabrera (changes requested):

src/stories/components/reset-button/reset-button.tsx
alejandrocabriales: ok, I will make the change

src/stories/components/cu-activity-table/cu-activity-table.placeholder.tsx
alejandrocabriales: You are right, in next refactoring, we will extend the component to have both functions

PR #391, opened 2022-11-03
Add InitialProps in _document y SSR in all client side

alejandrocabriales:
# Ticket (Any problem see this article)
https://dev.to/shinshin86/workaround-reference-the-cookie-value-in-document-js-in-next-js-32am
Analytic 

# Actions
Add InitialProps in _document y SSR in all client side


z3phyro (approved):
LGTM

PR #394, opened 2022-11-03
[NO-TICKET] Fix on transparency report comments filtering

z3phyro:


z3phyro:
some issue with github

PR #420, opened 2022-11-21
SES connect API delete account

alejandrocabriales:
# Ticket
SES-189-feature-useraccounts-v1

# Description

# Actions
Add delete function

Add back in Close button
Fix enable disable and go to profile
Fix router enable each profile
Add function helper to notification]
Move logic to mvvm file
Correct links to page
Add correct way to show user role


src/stories/containers/auth/change-password/change-password.tsx
```diff
@@ -1,17 +1,23 @@
...
 import { CircleAvatar } from '../../../components/circle-avatar/circle-avatar';
+import { useRouter } from 'next/router';
 
 export default () => {
   const { isLight } = useThemeContext();
+  const router = useRouter();
   const { form, username, loading, error } = userChangePasswordMvvm();
 
+  const handleGoBack = useCallback(() => {
+    router.back();
```
YasielCabrera, line 18: When the user goes directly to the URL, the "back" status will be not available so it will no able to go back. We can just redirect to the profile page maybe or check if the router has the possibility to go back and if it hasn't, just do the redirect

src/stories/containers/users/create-account/create-account.tsx
```diff
@@ -1,17 +1,23 @@
...
 import { useCreateAccountMvvm } from './create-account.mvvm';
+import { useRouter } from 'next/router';
 
 export default () => {
   const { isLight } = useThemeContext();
+  const router = useRouter();
   const { form, loading, error } = useCreateAccountMvvm();
 
+  const handleGoBack = useCallback(() => {
+    router.back();
```
YasielCabrera, line 18: The same comment of the previous `handleGoBack` function applies here. Maybe the functionality can be reused in both places and any other place where it can be useful

src/stories/containers/users/delete-account/delete-account.tsx
```diff
@@ -14,57 +14,83 @@ import { toast } from 'react-toastify';
...
+  const { clientRequest, user } = useAuthContext();
   const { userName, id } = router.query;
   const [value, setValue] = useState('');
   const { isLight } = useThemeContext();
   const handleChange = useCallback((value: React.ChangeEvent<HTMLInputElement>) => {
     setValue(value.target.value);
   }, []);
 
+  const handleGoBack = useCallback(() => {
+    router.back();
```
YasielCabrera, line 32: same thing that the previous `goBack` functions

src/stories/containers/users/delete-account/delete-account.tsx
```diff
@@ -14,57 +14,83 @@ import { toast } from 'react-toastify';
...
+              autoClose: 3000,
+              hideProgressBar: true,
+              closeButton: false,
+            }
+          );
+        }
       }
     } catch (error) {
-      console.log('error', error);
+      console.log({ value });
```
YasielCabrera, line 75: is this `console.log` needed?

src/stories/containers/users/delete-account/delete-account.tsx
```diff
@@ -14,57 +14,83 @@ import { toast } from 'react-toastify';
...
-            closeButton: false,
-          }
-        );
+      const { query: gqlQueryLogin, input } = LOGIN_REQUEST(user?.username || '', value);
+      const response = await request(GRAPHQL_ENDPOINT, gqlQueryLogin, input);
+      if (response) {
+        const { query: gqlQuery, filter } = USERS_DELETE_FROM_ADMIN(id as string);
+        const data = await clientRequest?.request(gqlQuery, filter);
+        if (data.userDelete) {
+          toast(
```
YasielCabrera, line 43: maybe we can create a `toast` helper function to trigger this one with the values so we can reuse it and avoid repeating things like color, positions, icons, etc. This will be very handy in this feature and future features that have this kind of toasts

src/stories/containers/users/delete-account/delete-account.tsx
```diff
@@ -14,57 +14,83 @@ import { toast } from 'react-toastify';
...
+        {
+          position: toast.POSITION.BOTTOM_CENTER,
+          autoClose: 3000,
+          hideProgressBar: true,
+          closeButton: false,
+        }
+      );
     }
-  }, [clientRequest, id, router]);
+  }, [clientRequest, id, router, user?.username, value]);
```
YasielCabrera, line 93: We should move most of this logic to an mvvm hook to follow the same pattern across the app

src/stories/containers/users/enable-disable-accounts/enable-disable-accounts.tsx
```diff
@@ -20,22 +20,31 @@ const resultPassword = fill(arrayPassword, 'a');
...
-    if (data) {
-      setChecked(data.userSetActiveFlag[0].active);
+    try {
+      const { query: gqlQuery, input } = ENABLE_DISABLE_USER_REQUEST(!checked, id as string);
+      const data = await clientRequest?.request(gqlQuery, input);
+      if (data) {
+        setChecked(data.userSetActiveFlag.active);
+      }
+    } catch (error) {
+      console.log('error', error);
```
YasielCabrera, line 36: is this `console.log` needed?

src/stories/containers/users/enable-disable-accounts/enable-disable-accounts.tsx
```diff
@@ -20,22 +20,31 @@ const resultPassword = fill(arrayPassword, 'a');
...
-  }, [checked, clientRequest]);
+  }, [checked, clientRequest, id]);
 
   const handleDeleteAccount = useCallback(() => {
     router.push('/auth/delete-account');
   }, [router]);
 
+  const handleGoBack = useCallback(() => {
+    router.back();
+  }, [router]);
```
YasielCabrera, line 46: We should move most of this login to an mvvm hook too

YasielCabrera (changes requested):

src/stories/containers/users/delete-account/delete-account.tsx
alejandrocabriales: You are right I will make the change

src/stories/containers/users/create-account/create-account.tsx
alejandrocabriales: You are right I will make the change

src/stories/containers/users/delete-account/delete-account.tsx
alejandrocabriales: No, my bad, I will remove it

src/stories/containers/users/delete-account/delete-account.tsx
alejandrocabriales: Already done

src/stories/containers/users/delete-account/delete-account.tsx
alejandrocabriales: You are right, I will move all the logic

src/stories/containers/users/enable-disable-accounts/enable-disable-accounts.tsx
alejandrocabriales: Just for check it , but it can be delete

src/stories/containers/users/enable-disable-accounts/enable-disable-accounts.tsx
alejandrocabriales: ok, thanks for the tips , already move

src/core/hooks/useIsAdmin.ts
```diff
@@ -0,0 +1,7 @@
+import { UserDTO } from '../models/dto/auth.dto';
+
+export const useIsAdmin = (user: UserDTO) => {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const isAdmin = (user as any)?.roles[0].name === 'SuperAdmin';
```
YasielCabrera, line 5: according to this, the user can be empty or it is possible that it doesn't have the `roles` property, so you should make sure that the `roles` property is actually an array, has at least 1 element and that element has the `name` property

Could we have an enum with the possible roles?

src/core/utils/string.utils.ts
```diff
@@ -54,19 +54,26 @@ export const getWalletWidthForWallets = (wallets: BudgetStatementWalletDto[]) =>
...
-      return 'Site Admin';
-    case 'User':
-      return 'User';
-    default:
-      return '';
+export const getCorrectRoleApi = (user: UserDTO) => {
+  const allPermission: string[] = [];
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  (user as any).roles.forEach((role: UserRole) => {
```
YasielCabrera, line 61: if the `UserDTO` can have a `roles` property, is not better to add that to the dto instead of type cast the object so we can utilize the typescript checks here?

YasielCabrera (changes requested):

src/core/hooks/useIsAdmin.ts
alejandrocabriales: Yes we can move the role to Enum, I will fix the useIsAdmin, because the way to determine that is through the permissions, I will make the change

src/core/utils/string.utils.ts
alejandrocabriales: You are right, lets move the roles to DTO

YasielCabrera (approved):

PR #422, opened 2022-11-22
Fix the problem in the Contextual menu

alejandrocabriales:
# Ticket

# Actions
Fix the problem in the Contextual menu


YasielCabrera (approved):
LGTM

PR #423, opened 2022-11-23
Show some page only if is admin

alejandrocabriales:
# Ticket

# Description

# Actions
Show some page only if is admin

### DIFF
--- a/src/stories/containers/users/user-profile/user-profile.tsx
+++ b/src/stories/containers/users/user-profile/user-profile.tsx
@@ -22,7 +22,7 @@ const resultPassword = fill(arrayPassword, 'a');
 const UserProfile = () => {
   const router = useRouter();
   const { isLight } = useThemeContext();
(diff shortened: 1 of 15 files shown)


src/stories/containers/users/user-profile/user-profile.tsx
```diff
@@ -35,6 +35,13 @@ const UserProfile = () => {
   }, [checked, clientRequest]);
 
   const handleDeleteAccount = useCallback(() => {
+    router.push({
+      pathname: '/auth/delete-account/',
+      query: {
+        userName: user?.username,
+        id: user?.id,
```
YasielCabrera, line 42: You're expecting here that the `user` can be empty/undefined, but if that is possible you should handle that somehow or you will get an error in the `delete-account` page due the user is undefined

alejandrocabriales: The way that is handle, is that the user doesn't exist, then will show a error when you try to delete the user

PR #430, opened 2022-11-25
[SES-189] Fix issues

YasielCabrera:
## Ticket

## Actions
- create an auth fetcher to use when the request requires authentication and avoid code duplication
- added a black border to action buttons in small resolutions
- No validate old password in the Change Password form
- Use the Wrong Old Password error message when the old password is incorrect
- Go back instead home when the password is changed successfully
- Fix the avatar in the Change Password page

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,6 @@
 {
     "cSpell.words": [
+        "Formik",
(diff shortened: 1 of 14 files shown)


alejandrocabriales (approved):
LGTM

PR #450, opened 2022-11-30
Ses issues manager account round 4

alejandrocabriales:
# Ticket

# Description

# Actions
Add function close menu when user scrolling

Move logic inside of MVVM in theme mode and add close when scrolling

Fix order in delete account


YasielCabrera (approved):
LGTM

PR #458, opened 2022-11-30
Correct when text is to long user name

alejandrocabriales:
Fix align when user name is too long

# Ticket
Fix size of user name

# Description

# Actions

Correct when text is to long user name


YasielCabrera (approved):
LGTM

PR #460, opened 2022-12-01
[FIX-189] Fix 404 message flicker on auth related routes on refresh

YasielCabrera:
## Ticket

## Actions
- add concept of `protected` routes to the project
- Redirect to login when the users try to enter to a protected route and they're no authenticated
- Show a forbidden message when the user are authenticated but they're not admins


alejandrocabriales (approved):
LGTM

PR #462, opened 2022-12-01
SES issues manager account round 4

alejandrocabriales:
# Ticket
UTA issues

# Actions
Fix-area-click-menu-theme-mode
Fix align and hover
Fix some issue relate padding

### DIFF
--- a/src/stories/containers/users/delete-account/delete-account.tsx
+++ b/src/stories/containers/users/delete-account/delete-account.tsx
@@ -71,8 +71,6 @@ const DeleteAccount: React.FC<{ username?: string }> = ({ username }) => {
           <Label isLight={isLight}>{`${
             isAdmin && !isAdminRouter
               ? 'Enter your admin password to delete the account'
-              : isAdmin && isAdminRouter
-              ? 'Enter password to delete the account'
               : 'Enter password to delete the account'
           }`}</Label>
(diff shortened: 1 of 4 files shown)


src/stories/containers/users/delete-account/delete-account.tsx
```diff
@@ -68,7 +69,11 @@ const DeleteAccount: React.FC<{ username?: string }> = ({ username }) => {
           }}
         >
           <Label isLight={isLight}>{`${
-            isAdmin ? 'Enter your admin password to delete the account' : 'Enter password to delete the account'
+            isAdmin && !isAdminRouter
+              ? 'Enter your admin password to delete the account'
+              : isAdmin && isAdminRouter
+              ? 'Enter password to delete the account'
+              : 'Enter password to delete the account'
```
YasielCabrera, line 76: Please, note that this is redundant as the message in both cases is the same one

YasielCabrera (changes requested):

src/stories/containers/users/delete-account/delete-account.tsx
alejandrocabriales: You are right, I will remove

YasielCabrera (approved):
LGTM

PR #476, opened 2022-12-11
[SES-203] Expense report status indicator

YasielCabrera:
## Ticket
https://trello.com/c/CrTkQJV6/203-feature-coreunitfinancestransparency-v4

## Description
Added the expense report status indicator but still pending the link logic
Added pending stories for transparency auditor comments components
Aligned Storybook breakpoints with the ones used in the app


src/stories/containers/transparency-report/transparency-auditor-comments/budget-status-select.tsx
```diff
@@ -1,23 +1,32 @@
...
+    statuses.length > 0 ? statuses[0] : BudgetStatus.Draft
+  );
+
+  const selectStatusHandler = useCallback(
+    (status: BudgetStatus) => {
+      setSelectedStatus(status);
+      onChangeStatus?.(status);
+    },
+    [setSelectedStatus, onChangeStatus]
   );
```
dvegap95, line 30: The useCallback hook is pointless in this case since the result of the hook is called later from within an arrow function that is the one passed on as a component propprop. We should either remove the hook or modify it's argument to be passed directly as the onClick prop of the component

```suggestion
  const selectStatusHandler = (status: BudgetStatus) => {
      setSelectedStatus(status);
      onChangeStatus?.(status);
  };
```

.storybook/preview.jsx
```diff
@@ -13,6 +13,59 @@ export const parameters = {
     nextRouter: {
         Provider: RouterContext.Provider,
     },
+    viewport: {
+        viewports: {
+            table_375: {
```
dvegap95, line 18: Typo here, maybe we can use "tablet" instead of "table"

dvegap95 (changes requested):

dvegap95 (approved):

PR #491, opened 2022-12-21
Fix: CU moving back when changing the budget month

YasielCabrera:
## Ticket
[No ticket]

## Description 
Fixed issue in the CU report tabs when changing tabs, the CU was moving to the previous CU


alejandrocabriales (approved):
LGTM

PR #495, opened 2022-12-29
SES add cookies dark mode

alejandrocabriales:
# Ticket
SES-159-qa-bug-findings-v-30-r

# Description

# Actions
Get isLight correct way

Update ThemeMode and Theme-Provider from cookies server
Update banner policy from cookie server
Update page delete conditional if
Update _app file

### DIFF
--- a/pages/_app.tsx
+++ b/pages/_app.tsx
@@ -19,7 +19,8 @@ import { AuthContextProvider } from '../src/core/context/AuthContext';
 import { getAuthFromStorage } from '../src/core/utils/auth-storage';
 import { ContainerNotification } from '../src/stories/components/notification/notification';
 import AppLayout from '../src/stories/containers/layout/layout';
-import { processCookieValue } from '../src/core/utils/cookie-helpers';
+import { parseCookie } from '../src/core/utils/cookie-helpers';
+import { CookiesInterface } from '../src/core/utils/types-utils';
 
 // disable useLayoutEffect SSR warnings to avoid log spamming the console
 // https://gist.github.com/gaearon/e7d97cdf38a2907924ea12e4ebdf3c85
@@ -29,26 +30,20 @@ export type NextPageWithLayout = NextPage & {
   getLayout?: (page: ReactElement) => ReactNode;
 };
 
-interface Cookies {
-  themeTracking: string;
-  timestampTracking: string;
-  analyticsTracking: string;
-}
 interface MyAppProps extends AppProps {
   Component: NextPageWithLayout;
   emotionCache?: EmotionCache;
   protected?: boolean;
   isLight: boolean;
-  cookiesObject: Cookies;
+  cookiesObject: CookiesInterface;
 }
 
 function MyApp(props: MyAppProps) {
   const { Component, pageProps, isLight, cookiesObject } = props;
-
   const router = useRouter();
 
   useEffect(() => {
-    if (gtag.GA_TRACKING_ID && cookiesObject.analyticsTracking === 'true') {
+    if (gtag.GA_TRACKING_ID && cookiesObject.allowsAnalyticsTracking) {
       const handleRouteChange = (url: URL) => {
         gtag.pageView(url);
       };
@@ -57,7 +52,7 @@ function MyApp(props: MyAppProps) {
         router.events.off('routeChangeComplete', handleRouteChange);
       };
     }
-  }, [cookiesObject.analyticsTracking, router.events]);
+  }, [cookiesObject.allowsAnalyticsTracking, router.events]);
 
   useEffect(() => {
     const authData = getAuthFromStorage();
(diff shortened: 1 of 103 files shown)


pages/_app.tsx
```diff
@@ -78,4 +90,27 @@ function MyApp(props: MyAppProps) {
...
+    analyticsTracking: '',
+  };
+
+  if (ctx.req?.headers.cookie) {
+    const decodedCookie = decodeURIComponent(ctx?.req?.headers.cookie || '');
+    const themeModeCookie = decodedCookie?.split(';');
+    cookiesObject.themeTracking = processCookieValue(themeModeCookie[0]);
+    cookiesObject.timestampTracking = processCookieValue(themeModeCookie[1]);
+    cookiesObject.analyticsTracking = processCookieValue(themeModeCookie[2]);
+    themeMode = processCookieValue(themeModeCookie[3]);
```
YasielCabrera, line 107: You should parse the cookies raw values and later get the values by cookies name, using the positions of the result could end up with unexpected errors as the ordering could not be the same every time, some cookies could be missing or we could add some others in the future. I recommend to use a solution like the follow one:

```js
const parseCookie = str =>
  str
  .split(';')
  .map(v => v.split('='))
  .reduce((acc, v) => {
    acc[decodeURIComponent(v[0].trim())] = decodeURIComponent(v[1].trim());
    return acc;
  }, {});
```
[source](https://www.geekstrick.com/snippets/how-to-parse-cookies-in-javascript/)

pages/_app.tsx
```diff
@@ -28,18 +29,26 @@ export type NextPageWithLayout = NextPage & {
   getLayout?: (page: ReactElement) => ReactNode;
 };
 
+interface Cookies {
+  themeTracking: string;
+  timestampTracking: string;
+  analyticsTracking: string;
```
YasielCabrera, line 35: Using boolean types instead of strings will make the work with those values easier as they only can be `'true'` or `'false'`. You just need to convert them to boolean when you read the cookie values. Changing the name to something like `allowThemeTracking` for example, would be more readable too

src/core/context/CookiesContext.tsx
```diff
@@ -20,7 +20,19 @@ export type CookiesContextValues = {
 const CookiesContextTracking = createContext<CookiesContextValues>({} as CookiesContextValues);
 const useCookiesContextTracking = () => useContext(CookiesContextTracking);
 
-const CookiesProviderTracking: React.FC<React.PropsWithChildren> = ({ children }) => {
+interface Props {
+  timestampTracking: string;
+  themeTracking: string;
+  analyticsTracking: string;
```
YasielCabrera, line 26: here you should use boolean types too

src/core/context/CookiesContext.tsx
```diff
@@ -20,7 +20,19 @@ export type CookiesContextValues = {
...
+  children: ReactNode;
```
YasielCabrera, line 27: This is not needed as you're using `React.PropsWithChildren & props` in the type declaration if the provider. A better solution if you use an `interface` would be:

```js
interface Props extends React.PropsWithChildren {
  // ...
}
```

src/stories/containers/cookies-policy/cookies-policy-banner.mvvm.ts
```diff
@@ -1,18 +1,24 @@
 import { useCallback, useMemo, useState } from 'react';
 import { useCookies } from 'react-cookie';
 import { useScrollLock } from '../../../core/hooks/scroll-hooks';
-import { useMediaQuery } from '@mui/material';
+interface Props {
+  timestampTracking: string;
+  themeTracking: string;
+  analyticsTracking: string;
```
YasielCabrera, line 7: here you should reuse the same type declared at the `_app.ts`, maybe move it to a type file like the `types_utils` or some other place where it can be reuser more easily as this type can be used in several places

YasielCabrera (changes requested):

src/stories/containers/cookies-policy/cookies-policy-banner.mvvm.ts
alejandrocabriales: done

src/core/context/CookiesContext.tsx
alejandrocabriales: ok

src/core/context/CookiesContext.tsx
alejandrocabriales: done

pages/_app.tsx
alejandrocabriales: OK, i will rename , and add boolean type

pages/_app.tsx
alejandrocabriales: You are right , __ will use this approach

YasielCabrera (approved):
LGTM

PR #496, opened 2023-01-05
Feature/203 improve last visit indicator

YasielCabrera:
## Ticket
https://trello.com/c/CrTkQJV6/203-feature-coreunitfinancestransparency-v4

## Description
If there are no logged users, the last indicator in the Activity Feed is "calculated" using data from local storage. If the user is logged, then the data is fetched from the API. The same behavior is considered for the comments in the CU budget statements.

## Actions
- improved the PermissionManager to store the JWT token of the logged user 
- Rewrite the last visit handler to use a local storage or API strategy depending if the user is logged or not
- Use the new logic in the Activity fees pages
- Fixed new comments indicator in the CU expense report budget statements


src/core/context/CommentActivityContext.tsx
```diff
@@ -0,0 +1,10 @@
+import { createContext, useContext } from 'react';
```
dvegap95, line 1: This way of importing/using react module elements seems inconsistent with the one used on other places in the code i.e.[React.useMemo instead of just useMemo](https://github.com/makerdao-ses/ecosystem-dashboard/pull/496/files#diff-6177c1c9dbecf396a26ad5b3eaf7180aa173734b7b8492b117c97feed111376fL38). Since React object is global it'll make no difference technically(IMHO) but maybe it's worth changing it in one of the places for the sake of consistency

src/stories/containers/transparency-report/transparency-report.mvvm.ts
```diff
@@ -249,13 +267,82 @@ export const useTransparencyReportViewModel = (coreUnit: CoreUnitDto) => {
...
+
+  // update the visit date (preventing multiple renderings)
+  let timeout: NodeJS.Timeout;
+  useEffect(() => {
+    clearTimeout(timeout);
+    timeout = setTimeout(async () => {
+      if (isTimestampTrackingAccepted) {
+        await lastVisitHandler.visit();
+      } else {
+        alert('hello');
```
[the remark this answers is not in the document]
YasielCabrera: Yes, it was a mistake. Will be fixed

src/core/context/CommentActivityContext.tsx
YasielCabrera: You're right, we should ensure consistency here, we should be adding a eslint rule for that. The way is done here should be the one used in the entire project since `React.useXXX` is more verbose and just using `useXXX` is enough as this is the convention

dvegap95 (approved):

PR #500, opened 2023-01-13
Fix-issue-relate-cookies

alejandrocabriales:
# Ticket
https://trello.com/c/vl1Dc2j6/222-feature-cookie-issue

# Actions
Add theme Mode cookies object and add expires date for all cookies

### DIFF
--- a/pages/_app.tsx
+++ b/pages/_app.tsx
@@ -94,20 +94,10 @@ MyApp.getInitialProps = async ({ ctx }: { ctx: NextPageContext }) => {
     const cookiesParsed = parseCookie(ctx.req?.headers.cookie);
 
     cookiesObject = {
-      ...cookiesObject,
       allowsThemeTracking: Boolean(cookiesParsed?.themeTracking),
-    };
-    cookiesObject = {
-      ...cookiesObject,
       allowsTimestampTracking: Boolean(cookiesParsed?.timestampTracking),
-    };
-    cookiesObject = {
-      ...cookiesObject,
       allowsAnalyticsTracking: Boolean(cookiesParsed?.analyticsTracking),
-    };
-    cookiesObject = {
-      ...cookiesObject,
-      themeModeCookie: cookiesParsed.themeModeCookie,
+      themeModeCookie: cookiesParsed?.themeModeCookie || 'light',
     };
   }
 
--- a/pages/_app.tsx
+++ b/pages/_app.tsx
@@ -83,21 +83,36 @@ function MyApp(props: MyAppProps) {
 }
 
 MyApp.getInitialProps = async ({ ctx }: { ctx: NextPageContext }) => {
-  let themeMode = '';
-  const cookiesObject: CookiesInterface = {
+  let cookiesObject: CookiesInterface = {
     allowsThemeTracking: false,
     allowsTimestampTracking: false,
     allowsAnalyticsTracking: false,
+    themeModeCookie: 'light',
   };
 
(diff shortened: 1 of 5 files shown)


pages/_app.tsx
```diff
@@ -83,21 +83,35 @@ function MyApp(props: MyAppProps) {
...
+      allowsTimestampTracking: Boolean(cookiesParsed?.timestampTracking),
+    };
+    cookiesObject = {
+      ...cookiesObject,
+      allowsAnalyticsTracking: Boolean(cookiesParsed?.analyticsTracking),
+    };
+    cookiesObject = {
+      ...cookiesObject,
+      themeModeCookie: cookiesParsed.themeModeCookie,
+    };
```
YasielCabrera, line 111: Please, note that you can do all this at the same time:
```js
cookiesObject = {
      allowsThemeTracking: Boolean(cookiesParsed?.themeTracking),
      allowsTimestampTracking: Boolean(cookiesParsed?.timestampTracking),
      allowsAnalyticsTracking: Boolean(cookiesParsed?.analyticsTracking),
      themeModeCookie: cookiesParsed?.themeModeCookie || 'light',
    };
```

src/core/utils/date.utils.ts
```diff
@@ -1,2 +1,8 @@
+import { DateTime } from 'luxon';
+
 export const API_MONTH_TO_FORMAT = 'yyyy-MM-01';
 export const API_MONTH_FROM_FORMAT = 'yyyy-MM-dd';
+
+// Expires date of cookies
+export const daysToExpire = new Date(DateTime.now().plus({ months: 10 }).toString());
+export const daysToDeleteCookie = new Date(DateTime.now().minus({ year: 2 }).toString());
```
YasielCabrera, line 8: This should be generated dynamically with a function, otherwise, the value will remain the same till the server is restarted or a release is made because the values are in memory till that. Note that this doesn't happen or it is not noted on development because the server reloads the values on every change (hot reloading)

Also note, that if you're using `luxon` dates, you can do just `DateTime.utc().plus({ months: 10 }).toISO()` instead of creating a native date to parse it again.

YasielCabrera (changes requested):

src/stories/containers/cookies-policy/cookies-policy-banner.mvvm.ts
```diff
@@ -48,25 +49,58 @@ export const useCookiesPolicyBannerMvvm = ({ cookiesObject }: Props) => {
...
+    removeCookie('analyticsTracking', {
+      expires: daysToDeleteCookie(),
+      path: '/',
+    });
+  }, [removeCookie]);
+
+  const deletedThemeCookie = useCallback(() => {
+    removeCookie('themeModeCookie', {
+      expires: daysToDeleteCookie(),
+      path: '/',
```
YasielCabrera, line 102: I found these ways to remove the cookies a bit hard to catch as the date to be deleted is 2 years (ago). Did you try to remove the `expires`? Like this: https://stackoverflow.com/a/69792708/11429218

Note that this is done on the client side, so removing the cookies with the standard library should also work.

If removing the `expires` (but keeping the `path` doesn't work, maybe we should try to use the `max-age` property: look here and example (in SSR) but the principle is the same: https://dev.vamsirao.com/clearing-cookie-in-nextjs

This way is more readable and easy to understand

YasielCabrera (changes requested):

YasielCabrera (approved):
LGTM

PR #501, opened 2023-01-13
SES 224 feature more obvious off boarded core unit UI

alejandrocabriales:
# Ticket
https://trello.com/c/BV8oP96F/224-feature-more-obvious-off-boarded-core-unit-ui

# Description

# Actions
Change fontSize and LineHeigth default component
Add correct default behavior

### DIFF
--- a/src/stories/containers/cu-table/cu-table-2.mvvm.ts
+++ b/src/stories/containers/cu-table/cu-table-2.mvvm.ts
@@ -1,4 +1,6 @@
 import request from 'graphql-request';
+import groupBy from 'lodash/groupBy';
+import orderBy from 'lodash/orderBy';
 import { DateTime } from 'luxon';
 import { useRouter } from 'next/router';
 import { useCallback, useMemo, useState } from 'react';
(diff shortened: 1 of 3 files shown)


src/stories/containers/cu-table/cu-table-2.mvvm.ts
```diff
@@ -211,13 +213,41 @@ export const useCoreUnitsTableMvvm = () => {
...
+      ? 3
+      : status === CuStatusEnum.Obsolete
+      ? 2
+      : status === CuStatusEnum.Rejected
+      ? 1
+      : 0;
+  const dataWithStatus = filteredData?.map((item) => ({
+    ...item,
+    status: giveWeightByStatus(getStautsMip39AccetedOrObsolete(item)),
+  }));
```
YasielCabrera, line 231: I'm not 100% sure, but `dataWithStatus` will be a new array on every render and it is used as a dependency in the `useMemo` of the next sentence so on every render the `useMemo` will be revalidating the value and doing the calculation. You should memorize this value too

YasielCabrera (changes requested):

src/stories/containers/cu-table/cu-table-2.mvvm.ts
alejandrocabriales: ok, you are right

YasielCabrera (approved):

PR #519, opened 2023-01-24
Hidden section when there is no data

alejandrocabriales:
#No ticket

#Action
Hidden section when there is no data


YasielCabrera (approved):
LGTM

PR #529, opened 2023-01-30
Add default order when the list is order

alejandrocabriales:
# Ticket
https://trello.com/c/YANCMcmX/231-qa-bug-findings-v-60

# Actions
Add default order when the list is order


YasielCabrera (approved):
LGTM

PR #543, opened 2023-02-02
Delimit range when min and max are close

alejandrocabriales:
# Ticket
https://trello.com/c/Ba6F5toO/215-qc-round-v-60-r

# Actions
Delimit range when min and max are close


YasielCabrera (approved):
LGTM

PR #567, opened 2023-02-08
[SES-NO_TICKET] Fix hidden line

YasielCabrera:
## Ticket
No ticket

## Description
Show the divider line on the Quarter carousel when there is just the exact number of quarters than the amount that need to be shown


alejandrocabriales:
LGTM

PR #573, opened 2023-02-09
Fix-month-name

alejandrocabriales:
## No Ticket

##Description:
Same number of character in each month

##Actions
Fix-month-name


YasielCabrera (approved):
LGTM

PR #599, opened 2023-02-20
[SES-QC-243] Fix body overflow when route changes using the bar-select component

YasielCabrera:
## Ticket
https://trello.com/c/nHLbSsUd/243-qc-round-v-70-r

## Description
Fix the scroll when the pages are being navigated using the top-bar-select component


alejandrocabriales (approved):
LGTM

PR #606, opened 2023-02-22
[SES-234] Use login modal instead of page on mobile devices

YasielCabrera:
## Ticket
https://trello.com/c/YzBgSTpH/234-feature-useraccounts-v2

## Description
The login page should not be shown on mobile devices, in that case, the login modal is the one that should be shown.


alejandrocabriales (approved):
LGTM

PR #644, opened 2023-03-02
Hotfix/hide del core unit

YasielCabrera:
## Ticket
No ticket

## Description
Hide DEL core unit from the Core units list as it is not a core unit


alejandrocabriales (approved):
LGTM

PR #646, opened 2023-03-03
Issues/fix jump navigation component

alejandrocabriales:
# Ticket
https://trello.com/c/NaQQSUtv/254-qc-round-v-80-r

#What solved
 Expense Reports and Recognized Delegates views. Should maintain the position of the component when navigating through the months. 

# Actions
Add height to container to avoid the jump
Fix correct jump in transparency page


YasielCabrera (approved):
LGTM

PR #661, opened 2023-03-09
Add tooltip desktop and show in the icon

alejandrocabriales:
# Ticket

# What solved
Should add an information tooltip on the "Target" column (ArrowToolTip, TransferTargetToolTip)
Should open the tool tip when the user clicks on the information icon (mobile) 

# Actions
Add tooltip desktop and show in the icon


src/stories/components/ArrowPopover/ArrowPopover.tsx
```diff
@@ -0,0 +1,15 @@
+import styled from '@emotion/styled';
+import React from 'react';
+
+import type { PropsWithChildren } from 'react';
+type Props = PropsWithChildren;
```
YasielCabrera, line 5: If the props are not going to add any other prop, just use `PropsWithChildren` directly

YasielCabrera (changes requested):

src/stories/components/ArrowPopover/ArrowPopover.tsx
alejandrocabriales: for now just children

YasielCabrera (approved):
LGTM

PR #666, opened 2023-03-10
Ses 245 feature coreunittransparencyreporting v6 add fiels table

alejandrocabriales:
# Ticket

#What solved
Under the existing Transfer Requests Tab, change the column text: "Current Balance” → “31-JAN Balance". This is specific to the expense report, and is derived from the specific data that the report was published. 

# Actions
Add headers to table
Add new style for data in the column


YasielCabrera (approved):
LGTM

PR #669, opened 2023-03-15
Show mobile modal-sheet when is real device

alejandrocabriales:
# Ticket

#What solved
 Should create the `OverlayedTooltip` for actual mobile devices 
 Show mobile modal-sheet when is real device

# Actions
Show mobile modalsheet when is real device


YasielCabrera (approved):
LGTM

PR #676, opened 2023-03-20
[No-ticket] Position of dot 375 view

alejandrocabriales:
# Ticket
No ticket

#What solved
Position the dot in the quarterly card

# Actions
Position of dot 375 view

### DIFF
--- a/src/stories/components/CookiesPolicyBanner/CookiesPolicyBanner.tsx
+++ b/src/stories/components/CookiesPolicyBanner/CookiesPolicyBanner.tsx
@@ -1,4 +1,5 @@
 import styled from '@emotion/styled';
+import { useMediaQuery } from '@mui/material';
 import { useThemeContext } from '@ses/core/context/ThemeContext';
 import React from 'react';
 import lightTheme from '../../../../styles/theme/light';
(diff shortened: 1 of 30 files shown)


src/stories/components/CookiesPolicyBanner/CookiesPolicyBanner.tsx
```diff
@@ -60,6 +62,7 @@ const CookiesPolicyBanner = ({
               width: 285,
               height: 48,
               padding: '14.5px 76px',
+              marginRight: notMobile ? 24 : 0,
```
YasielCabrera, line 65: We should avoid this as much as possible use this approach and do a native solution instead:

You can do this:
```ts
const StyledButton = styled(CustomButton)({
  // your styled here...

  '& > div': {
    // ... your button text styles here...
  }

  [lightTheme.breakpoints.up('table_834')]: {
    marginRight: 24,
    // ...
  }
});
```

The previous snippet works because you can extend custom components that receive a `className` prop, behind scenes  `styled` transform you style in CSS style with a `className`

YasielCabrera (changes requested):

src/stories/components/CookiesPolicyBanner/CookiesPolicyBanner.tsx
alejandrocabriales: you are right let's move to that solution

YasielCabrera (approved):
LGTM

PR #679, opened 2023-03-21
Fix align card target balance column

alejandrocabriales:
# Ticket

#What solved
Should be aligned the info icon and the value with the Target Balance text. 

# Actions
Fix align card target balance column

### DIFF
--- a/src/stories/components/TransparencyCard/TransparencyCard.tsx
+++ b/src/stories/components/TransparencyCard/TransparencyCard.tsx
@@ -75,9 +75,7 @@ const Row = styled.div<{ hasIcon?: boolean; height?: string }>(({ hasIcon = fals
   display: 'flex',
   alignItems: hasIcon ? 'flex-start' : 'center',
   flex: 1,
-  '& > div > div > div:nth-child(2) > div >div:nth-child(2) > div': {
-    paddingTop: 4,
-  },
+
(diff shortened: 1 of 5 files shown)


src/stories/components/TransparencyCard/TransparencyCard.tsx
```diff
@@ -74,8 +74,10 @@ const FooterWrapper = styled.div<{ isLight: boolean }>(({ isLight }) => ({
 const Row = styled.div<{ hasIcon?: boolean; height?: string }>(({ hasIcon = false }) => ({
   display: 'flex',
   alignItems: hasIcon ? 'flex-start' : 'center',
-
   flex: 1,
+  '& > div > div > div:nth-child(2) > div >div:nth-child(2) > div': {
```
YasielCabrera, line 78: please, use another approach to add styles to that `div`, there's too much depth in that selector

YasielCabrera (changes requested):

src/stories/components/TransparencyCard/TransparencyCard.tsx
alejandrocabriales: you are right, the change are done

YasielCabrera (approved):
LGTM

PR #683, opened 2023-03-22
[SES-245] Connect Transfer-Request Tab with the API

alejandrocabriales:
# Ticket

#What solved
Should display the correct "target" value coming from the API in the "Target account" column
Should display the MIP number and the link coming from the API on the tooltip
Should open (https://mips.makerdao.com/mips/details/MIP40) in a new tab when the user clicks on the tooltip link 

# Actions
Add Dto and parameters to query

### DIFF
--- a/src/core/models/dto/coreUnitDTO.ts
+++ b/src/core/models/dto/coreUnitDTO.ts
@@ -75,9 +75,23 @@ export interface BudgetStatementLineItemDto {
   group?: string;
 }
 
+export interface SourceDto {
+  code: string;
+  url: string;
+  title: string;
+}
+export interface TargetDto {
+  amount: number;
+  calculation: string;
+  description: string;
+  source: SourceDto;
+}
+
 export interface BudgetStatementWalletTransferRequestDto {
   requestAmount: number;
   walletBalance: number;
+  target: TargetDto;
+  walletBalanceTimeStamp: string;
 }
 
 export interface BudgetStatementWalletDto {
--- a/src/stories/containers/TransparencyReport/transparencyReportAPI.ts
+++ b/src/stories/containers/TransparencyReport/transparencyReportAPI.ts
@@ -64,8 +64,19 @@ export const CORE_UNIT_REQUEST = (shortCode: string) => ({
               payment
             }
             budgetStatementTransferRequest {
+              target {
+                amount
+                calculation
+                description
+                source {
+                  code
+                  url
+                  title
+                }
+              }
               requestAmount
               walletBalance
+              walletBalanceTimeStamp
             }
           }
           budgetStatementMKRVest {


src/stories/containers/TransparencyReport/components/TransparencyTransferRequest/useTransparencyTransferRequest.tsx
```diff
@@ -89,6 +98,44 @@ export const useTransparencyTransferRequest = (currentMonth: DateTime, budgetSta
...
+      return '';
+    }
+
+    const dateTime = DateTime.fromISO(timeStampAnyWallet);
+    const formatData = dateTime.toFormat('dd-LLL');
+
+    return formatData;
+  }, [wallets]);
+
+  const getTransferRequestTargetBalanceColumnTotal = useMemo(() => {
```
YasielCabrera, line 126: Suggestion: this name looks like a function but it is actually a number, maybe you should remove the `get` prefix

src/stories/containers/TransparencyReport/components/TransparencyTransferRequest/useTransparencyTransferRequest.tsx
```diff
@@ -89,6 +98,44 @@ export const useTransparencyTransferRequest = (currentMonth: DateTime, budgetSta
     return result;
   }, [currentMonth, budgetStatements]);
 
+  const getTransferRequestTargetBalanceColumn = useCallback((wallet: BudgetStatementWalletDto) => {
+    const targetWithTimeSpan: Pick<BudgetStatementWalletTransferRequestDto, 'target' | 'walletBalanceTimeStamp'> =
+      {} as BudgetStatementWalletTransferRequestDto;
+    wallet.budgetStatementTransferRequest?.forEach((item: BudgetStatementWalletTransferRequestDto) => {
+      targetWithTimeSpan.target = item.target;
+      targetWithTimeSpan.walletBalanceTimeStamp = item.walletBalanceTimeStamp;
+    });
```
YasielCabrera, line 107: In this `forEach` you're actually setting to `targetWithTimeSpan.target` and `targetWithTimeSpan.walletBalanceTimeStamp` the values of `target` and `walletBalanceTimeStamp` of the last `budgetStatementTransferRequest` item, is that what you actually want to do? if the answer is yes, this code is equivalent to:

```ts
const lastIndex = wallet.budgetStatementTransferRequest?.length - 1;
targetWithTimeSpan.target = wallet.budgetStatementTransferRequest?[lastIndex]?.target;
targetWithTimeSpan.walletBalanceTimeStamp = wallet.budgetStatementTransferRequest?[lastIndex]?.walletBalanceTimeStamp;
```

src/stories/components/TransparencyCard/TransparencyCard.tsx
```diff
@@ -19,24 +19,27 @@ export const TransparencyCard = (props: Props) => {
...
-              justifyContent: props.itemType ? 'flex-end' : undefined,
-              width:
-                header === 'Target Balance' || (header === 'Target Balance' && props.itemType !== 'total')
-                  ? '100%'
-                  : undefined,
-            }}
+        <>
+          <Row
+            key={header.toString()}
+            hasIcon={header !== 'Target Balance' || (header === 'Target Balance' && props.itemType === 'total')}
```
YasielCabrera, line 25: This looks kinda a patch, and in the future, more icons will be added to other columns, or the Column names could change using this approach will make really hard to change the names of the columns and keep it bugs free. Could you use any different approach to archive this?

YasielCabrera (changes requested):

src/stories/containers/TransparencyReport/components/TransparencyTransferRequest/useTransparencyTransferRequest.tsx
```diff
@@ -1,12 +1,21 @@
+/* eslint-disable no-lone-blocks */
+/* eslint-disable @typescript-eslint/no-unused-vars */
```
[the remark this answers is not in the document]
alejandrocabriales: It was my mistake, I made a commitment to delete those, but I  forgot to push. Done

src/stories/containers/TransparencyReport/components/TransparencyTransferRequest/useTransparencyTransferRequest.tsx
alejandrocabriales: I got you

src/stories/containers/TransparencyReport/components/TransparencyTransferRequest/useTransparencyTransferRequest.tsx
alejandrocabriales: You are right, in fact, this is only one item in the budgetStatementTransferRequest, so I got your point, and your approach

YasielCabrera: This has not been fixed yet

YasielCabrera (changes requested):

src/stories/containers/TransparencyReport/components/TransparencyTransferRequest/useTransparencyTransferRequest.tsx
alejandrocabriales: my bad, I take you approach but don't delete the old code

PR #685, opened 2023-03-23
Add new layout target balance

alejandrocabriales:
# Ticket

#What solved
Should display less gap between the numbers and the icon, and the header could be fully aligned with the values

# Actions
Add new layout target balance


YasielCabrera (approved):
LGTM

PR #688, opened 2023-03-24
[SES-246] Connect "By Category" section with the API

YasielCabrera:
# Ticket
https://trello.com/c/tqmW07KT/246-feature-aggregatedviewlevels-v1

# Description
Integrate the "By Category expense" section with the API

# What solved
- Should update the table values when the user change the filter value
- Should when "By Expense Category" is apply, separe by Headcount and Non - Headcount Expenses categories
- Should Headcount  Expenses will only be comprised of Compensation & Benefit, but extendes to next iteration be: But the pattern makes sense because in future iterations we will display the composition of Compensation & Benefits (ie… Smart Contract Development, Marketing, etc)
- Should Non Headcount: follow the same logic "By Budge": but for “Expense Category” it will be for the Top 7 Non Headcount Expenses, and the 8th will be the smallest out of the Top 8.
- Should  if there is 8 o more included  “All Remaining Non Headcount” if so , then will stay at that position, if not apperas should add and the end, in 8 position
- Should  the expense Expense Category column have The budget category column will list budget categories by name, under their specific sub header (“Headcount”, “Non Headcount” )


src/stories/containers/FinancesOverview/useFinancesOverview.ts
```diff
@@ -16,7 +16,9 @@ const fullBorder = [6, 6, 6, 6];
 const useFinancesOverview = (
   quarterExpenses: ExpenseDto[] = [],
   monthly: Partial<ExpenseDto>[],
-  breakdownExpenses: ExtendedExpense[]
+  byBudgetBreakdownExpenses: ExtendedExpense[],
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
alejandrocabriales, line 25: It's this disable of eslint necessary?

PR #708, opened 2023-04-05
[SES-257] Add navigable behavior to the tabs

YasielCabrera:
# Ticket

# Description
Added a button to the tabs components to compress/expand the tabs. Also, allow opening in a new browser tab the tabs components. The old tabs component was marked as a legacy for backward compatibility till the new implementation replaces it completely and it is deleted from the source code.

# What solved
- Should add a recognizable icon to the right of the "comments" tab
- Should display the tooltip "auditor view." when the user hovers over the icon
- Should replace the current tabs with just  "Budget Report" and "Comments." tabs when the "auditor view" button is clicked
- Should allow clicking the "auditor view" button and initiating the user interaction, replace the icon with a back icon that allows users to return to the default view. This feature ensures easy navigation between the auditor and default views as needed.
- Should restore the tabs when the "auditor view" button is clicked and the tabs are collapsed
- Should allow opening in a new tab when the user right click on the tabs


alejandrocabriales (approved):
LGTM

PR #710, opened 2023-04-05
[SES-257] Upgrade delegate's tabs

YasielCabrera:
# Ticket

# Description
Upgrade the tabs on the Delegates page to allow to open tabs in a new browser tab. Also optimized the code in the MVVM hook of the delegates page with the workflow for the current `Tabs` component

# What solved
- Should allow opening in a new browser tab when the user right click on the tabs on the Recognize delegates page


alejandrocabriales (approved):
LGTM

PR #711, opened 2023-04-06
[SES-257] Upgrade auth manage tabs

YasielCabrera:
# Ticket

# Description
The tabs under `/auth/manage` were upgraded to allow them to open in a new browser tab. Also the `Tabs` component was upgrade to allow it to be controlled/uncontrolled in a way that, when it is necessary the developer can control which tab is selected without updating the URL

# What solved
- Should allow opening in a new tab when the user right click on the "manage accounts" tabs


mact200590:
LGTM

PR #720, opened 2023-04-14
[SES-257] Manage tabs through query params

YasielCabrera:
# Ticket

# Description
Updated the Tabs component to be self-controlled using the URL query params instead of the hash, this way we can have several tabs on the same page without any conflict by using different query params keys. The query param key for all the current tabs was updated.

# What solved
- Should update the `Tabs` component to allow to store of the state of the selected tab through a query string instead of a hash
- Should allow to share the state of the selected tabs through the `section` query string for the main tabs, `actual-account` for the tabs of the actuals breakdowns tabs and `forecast-account` for the forecast breakdown tabs


alejandrocabriales (approved):
LGTM

PR #728, opened 2023-04-18
[SES-257] Improve Breakdown tabs behavior on Budget Report section

YasielCabrera:
# Ticket

# Description
Allow to compress/expand the breakdown tabs in the actuals and forecast sections simultaneously. The state of the tabs is shareable through the URL and the state should be kept after the page refresh or the URL opened on any other device.

# What solved
- Should compress the actuals and forecasts breakdown at the same time
- Should compress all the breakdown tabs by default when the auditor view is opened
- Should allow to share the state of the breakdown tabs (compressed/expanded) through the URL utilizing the `breakdownView` query string

### DIFF
--- a/src/core/utils/urls.ts
+++ b/src/core/utils/urls.ts
@@ -34,3 +34,13 @@ export const toAbsoluteURL = (relativeURL: string): string => {
   }
   return `${BASE_URL}${relativeURL}`;
 };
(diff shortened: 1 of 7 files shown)


alejandrocabriales (approved):
LGTM

PR #734, opened 2023-04-20
Feature/center years

alejandrocabriales:
# Ticket

# What solved
 ✅ Should add the delegate expense trend check for data for other views (Years should be center)

# Actions
Center the years in chart


YasielCabrera (approved):
LGTM

PR #743, opened 2023-04-27
Add SEO for delegate page

alejandrocabriales:
# Ticket

# What solved
✅Should add  SEO for the page
✅Should update the naming rules in the software to display the label "STRETCHED" in the budget graph on the core units index table when the actuals equal the budget cap, instead of the current "OVER BUDGET" label. 

# Actions
Add SEO for the delegate page
Add Seo and use copyIcon component
Add correct style when actuals and bugCap are the same


YasielCabrera (approved):
LGTM

PR #745, opened 2023-04-27
[SES-257] Added the values to the Mthly Budget column

YasielCabrera:
# Ticket

# Description
Added the real values to the Mthly Budget column

# What solved
- Should add a new column titled "MTHLY BUDGET" immediately to the right of the "wallet" column, which will display the budget cap for each entry.


alejandrocabriales (approved):
LGTM

PR #746, opened 2023-04-28
Add color gray and give hight when the value is cero

alejandrocabriales:
# Ticket

# What solved
✅Should The bars should display gray color when there is no value in the month

# Actions
Add the color gray and give height when the value is cero


YasielCabrera (approved):
LGTM

PR #752, opened 2023-05-05
[SES-265] Fix the delegates transparency reports page

YasielCabrera:
# Ticket
https://trello.com/c/Z1EJXde6/265-qc-round-v-90-r

# Description
The delegates transparency page got broken after the Delegates was removed from the CU endpoint as it is not a CU, the page was fixed using the new way: fetching directly the budget statements owned by the delegates

# What solved
-  Should fix the Recognize delegates report page after the API breaking change.
- Should update the spacing of the section: Actuals and Forecast of the Delegates repots reports page


alejandrocabriales (approved):
LGTM

PR #755, opened 2023-05-06
[SES-265] Fix Auditor view breaks on refresh

YasielCabrera:
# Ticket
https://trello.com/c/Z1EJXde6/265-qc-round-v-90-r

# Description
When the user navigates between the core units with the Auditor view activated and then refresh the page it breaks

# What solved
- When accessing the SH, COM, GROW, KING core units, an error is displayed (the access can be through the url or core unit navigators).


alejandrocabriales (approved):
LGTM

PR #765, opened 2023-05-09
Add disable in filter delegates

alejandrocabriales:
# Ticket

https://trello.com/c/kUYUN01m/266-qc-round-v-100-r

# What solved
✅Dark Mode. Filters in Core Units, Activity Feed, Delegates views. **Expected Output:** The hover should be displayed with color: #25273D. The last element of the area should be displayed as disabled.

# Actions
Add disable in filter delegates


YasielCabrera (approved):
LGTM

PR #767, opened 2023-05-10
Add hover in dark mode and delete unnecessary css

alejandrocabriales:
# Ticket

https://trello.com/c/kUYUN01m/266-qc-round-v-100-r

# What solved
✅Add hover in dark mode and delete unnecessary css

# Actions
Add hover in dark mode and delete unnecessary css

### DIFF
--- a/src/stories/containers/RecognizedDelegates/components/FilterDelegate/DelegateSelectItem.tsx
+++ b/src/stories/containers/RecognizedDelegates/components/FilterDelegate/DelegateSelectItem.tsx
@@ -80,9 +80,21 @@ const CheckWrapper = styled.span({
 const StyledContainer = styled(Container)<{ isLight: boolean; checked: boolean; isDisable: boolean }>(
   ({ isDisable, checked, isLight }) => ({
     '&:hover': {
-      background: isDisable ? 'none' : isLight ? (checked ? '#EDEFFF' : '#F6F8F9') : isDisable ? 'none' : '#25273D',
-      backgroundColor:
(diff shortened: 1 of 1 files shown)


src/stories/containers/RecognizedDelegates/components/FilterDelegate/DelegateSelectItem.tsx
```diff
@@ -80,9 +80,21 @@ const CheckWrapper = styled.span({
...
+          : !isDisable && checked
+          ? '#EDEFFF'
+          : '#F6F8F9'
+        : isDisable && checked
+        ? '#231536'
+        : isDisable && !checked
+        ? 'none'
+        : !isDisable && checked
+        ? '#231536'
+        : '#25273D',
```
YasielCabrera, line 97: As this has 3 different variables, it is pretty hard to follow, would be nice to create a variable in the styled component function to make it more readable. Maybe something like this:
```ts
const StyledContainer = styled(Container)<{ isLight: boolean; checked: boolean; isDisable: boolean }>(
  ({ isDisable, checked, isLight }) => {
  let background: string;
  if (isLight) {
    if (isDisabled) {
      background = checked ? 'xxx' : 'yyy';
    } else {
      background = checked ? 'qqq' : 'www';
    }
  } else {
    // ...
  }

  return ({
    '&:hover': {
      background,
    }
  })
});
```

alejandrocabriales: I got your point

YasielCabrera (approved):
LGTM

PR #782, opened 2023-05-23
Add center arrow when is mobile and not real device

alejandrocabriales:
# Ticket

# What solved
✅Should show green bars: Least important forecasts should be displayed as green bars with approximately 50% opacity. These bars represent situations where the forecast is well below the budget cap.(percent > 0 && percent <= 90)
✅Should show: When users hover over the dotted line of each forecast/budget cap bar, a tooltip should appear providing relevant information about that specific bar.
✅Should Add responsive and dark mode for the view

# Actions
Add a center arrow when is mobile and not a real device


YasielCabrera (approved):
lgtm

PR #789, opened 2023-05-26
Add correct color dark mode and fix arrow position

alejandrocabriales:
# Ticket

# What solved
✅The change was made but in dark mode, the text (Totals) is barely visible. Check the font color in dark mode.
✅ Resolution: 1194, the tooltip is cut off on the right side
✅ The tooltip should be displayed. It is a little bit difficult to see the modal. 

# Actions
Add the correct color dark mode and fix the arrow position


YasielCabrera (approved):
LGTM

PR #794, opened 2023-05-30
Fix popover don't show

alejandrocabriales:
# Ticket

# What solved
✅It is hard to touch the tooltip area
✅Hovering/clicking on the bars, the tooltip should be displayed. The touch area should be the actual number,  budget number, and bars. **Current Output:** The tooltip still missing. 
✅ Hovering over the 2 last values from the Totals column. **Expected Output:** The tooltip should be visible. **Current Output:** When the values are closer to the bottom, the tooltip is not displayed

# Actions
Fix popover don't show


YasielCabrera (approved):
LGTM

PR #797, opened 2023-05-30
Move popover when arrow have different position

alejandrocabriales:
# Ticket

# What solved
✅Example core unit. **Expected Output:**The tooltip should be displayed properly having into account the dots' position.  **Current Output:** The tooltip is displayed a little bit far from the dots.  

# Actions
Move popover when arrow have different position


YasielCabrera (approved):
LGTM

PR #800, opened 2023-05-31
Add condition hidden % values negative

alejandrocabriales:
# Ticket

# What solved

✅SAS core unit. Breakdown. **Expected Output:** In case of negative numbers, let’s just hide the % row. **Current Output:** The forecast and the percent are displayed with a negative value.  **Visual Proof:** 

# Actions

Add condition hidden % values negative


YasielCabrera (approved):
LGTM

PR #803, opened 2023-05-31
Delete the cero and percent when forecast and not budgets

alejandrocabriales:
# Ticket

# What solved
✅ **Expected Output:** Let’s just hide the % row and the 0 on the top right. **Current Output:** It's displayed a "-" in the percent and a 0 on the top right. 

# Actions
Delete the cero and percent when forecast and not budget cap


YasielCabrera (approved):
LGTM

PR #808, opened 2023-06-05
Feature/modal container modal UI

alejandrocabriales:
# Ticket
https://trello.com/c/8lQtM5Hz/276-user-story-finance-homepage-expense-categories-access

# What solved
✅ Should have the modal Description title
✅Should  the modal have a close (x) option at the top right

# Actions
Add property SVG and add Accordion for description 
Add builder and story for modal contained

### DIFF
--- a/src/core/businessLogic/builders/categoriesBuilders.ts
+++ b/src/core/businessLogic/builders/categoriesBuilders.ts
@@ -1,0 +1,16 @@
+export class CategoryBuilder {
+  private _category: string;
+
+  constructor() {
+    this._category = '';
+  }
+
+  withCategory(category: string): CategoryBuilder {
+    this._category = category;
+    return this;
+  }
+
+  build(): string {
+    return this._category;
+  }
+}
--- a/src/stories/components/BasicModal/CategoryItem/ArrowAccordionCategory.tsx
+++ b/src/stories/components/BasicModal/CategoryItem/ArrowAccordionCategory.tsx
@@ -25,17 +25,7 @@ const AccordionCategory: React.FC<Props> = ({ style, category, subCategory }) =>
       <Accordion>
         <AccordionSummary isLight={isLight}>{category}</AccordionSummary>
         <AccordionDetails>
-          <div
-            style={{
-              background: '#FBFBFB',
-              boxShadow: '0px 20px 40px rgba(219, 227, 237, 0.4), 0px 1px 3px rgba(190, 190, 190, 0.25)',
-              padding: 64,
-              borderRadius: '0px 0px 6px 6px',
-              textAlign: 'center',
-            }}
-          >
-            {subCategory}
-          </div>
+          <ItemsStyle>{subCategory}</ItemsStyle>
         </AccordionDetails>
       </Accordion>
     </TransactionHistoryContainer>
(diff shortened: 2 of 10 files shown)


src/core/businessLogic/builders/categoriesBuilders.ts
```diff
@@ -0,0 +1,16 @@
+export class CategoryBuilder {
```
YasielCabrera, line 1: Builders should be dedicated to complex object, in this case you just need a string so this should not needed

src/stories/components/BasicModal/CategoryItem/ArrowAccordionCategory.tsx
```diff
@@ -0,0 +1,99 @@
...
+        <AccordionDetails>
+          <div
+            style={{
+              background: '#FBFBFB',
+              boxShadow: '0px 20px 40px rgba(219, 227, 237, 0.4), 0px 1px 3px rgba(190, 190, 190, 0.25)',
+              padding: 64,
+              borderRadius: '0px 0px 6px 6px',
+              textAlign: 'center',
+            }}
+          >
```
YasielCabrera, line 36: This should be moved to a styled component in order to avoid inline styles

src/stories/containers/TransparencyReport/utils/descriptionsConstants.ts
```diff
@@ -0,0 +1,2 @@
+export const CANONNICAL_DESCRIPTION =
+  "Canonical Expense Categories in MakerDAO are standardized classifications of expenses tailored for managing MakerDAO's operational costs.";
```
YasielCabrera, line 2: This will probably be better to add it to the site it is needed as it is not reused

YasielCabrera (changes requested):

src/core/businessLogic/builders/categoriesBuilders.ts
alejandrocabriales: For now, its, just a string, but the object will increase , because it's a category with subcategories, so whan=t do you think to add the news properties as needed?

src/stories/containers/TransparencyReport/utils/descriptionsConstants.ts
alejandrocabriales: ok, i will

src/stories/components/BasicModal/CategoryItem/ArrowAccordionCategory.tsx
alejandrocabriales: done

src/core/businessLogic/builders/categoriesBuilders.ts
YasielCabrera: Sure, let's look forward and see how it evolves then

src/stories/components/BasicModal/ContainerModal.tsx
```diff
@@ -0,0 +1,170 @@
...
+              .map((item) => (
+                <CategoryItem category={item} />
+              ))}
+          </ContainerPar>
+          <ContainerOdd>
+            {noHeadCountCategories
+              ?.filter((_, index) => index % 2 !== 0)
+              .map((item) => (
+                <CategoryItem category={item} />
+              ))}
```
YasielCabrera, line 55: I strongly recommend implementing this using CSS columns in a way each category represents a column, it will be easier to implement responsive, to integrate with the API and more efficient than "manually" create the columns. You can take a look how bootstrap columns o MUI Grid works in columns

alejandrocabriales: You are right , done

PR #813, opened 2023-06-07
[SES-268] Integrate Snapshots with API

YasielCabrera:
# Ticket

# Description
Created the base for the integration of the Accounts Snapshots with the UI

# What solved
- Should fetch the snapshots information from the API in SSR
- Should return a 404 page when there is no snapshot data


alejandrocabriales (approved):
LGTM

PR #820, opened 2023-06-09
[SES-268] Add owner subtitle according to the data from the API

YasielCabrera:
# Ticket

# Description
Show the right subtitle according to the data coming from the API. Note that the general message is still to be defined.

# What solved
- Should create the right subtitle label according to the owner type requested

### DIFF
--- a/src/stories/containers/TransparencyReport/components/AccountsSnapshot/api/queries.ts
+++ b/src/stories/containers/TransparencyReport/components/AccountsSnapshot/api/queries.ts
@@ -39,7 +39,35 @@ export const accountsSnapshotQuery = (filter: SnapshotFilter) => ({
       }
     }
   `,
-  filter,
+  filter: {
+    filter,
+  },
+});
+
+export const coreUnitShortCodeQuery = (id: string) => ({
+  query: gql`
+    query CoreUnits($filter: CoreUnitFilter) {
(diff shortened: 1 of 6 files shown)


src/stories/containers/TransparencyReport/components/AccountsSnapshot/api/queries.ts
```diff
@@ -52,7 +80,57 @@ export const fetchAccountsSnapshot = async (ownerType: string, ownerId: string):
...
+interface TeamResponse {
+  teams: {
+    shortCode: string;
+  }[];
+}
+
+const getCoreUnitShortCode = async (ownerId: string): Promise<string> => {
+  const { query, filter } = coreUnitShortCodeQuery(ownerId);
+  const res = await request<CoreUnitResponse>(GRAPHQL_ENDPOINT, query, filter);
+  if (res && res.coreUnits && res.coreUnits[0] && res.coreUnits[0].shortCode) {
```
alejandrocabriales, line 101: In this case, it's not a better option to use. operator?

YasielCabrera: You're right, changing now...

alejandrocabriales (approved):
LGTM

PR #821, opened 2023-06-09
Fix/some conflictive stories

YasielCabrera:
# Ticket 
No ticket

# Description
Fix the `ItemCoreUnit` and `Activity Table` stores, they was coming up on chromatic on each PR without having any real change


alejandrocabriales (approved):
LGTM

PR #825, opened 2023-06-12
Fix format small category

alejandrocabriales:
# Ticket

https://trello.com/c/8lQtM5Hz/276-user-story-finance-homepage-expense-categories-access

# What solved
✅Fix the correct format for the subcaterory

# Description
This is an small Pr to fix the format of the category item

### DIFF
--- a/src/stories/components/BasicModal/CategoryItem/ArrowAccordionCategory.tsx
+++ b/src/stories/components/BasicModal/CategoryItem/ArrowAccordionCategory.tsx
@@ -5,6 +5,7 @@ import MuiAccordionSummary from '@mui/material/AccordionSummary';
 
 import { SelectChevronDown } from '@ses/components/svg/select-chevron-down';
 import { useThemeContext } from '@ses/core/context/ThemeContext';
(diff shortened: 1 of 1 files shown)


src/stories/components/BasicModal/CategoryItem/ArrowAccordionCategory.tsx
```diff
@@ -28,7 +29,7 @@ const AccordionCategory: React.FC<Props> = ({ style, category }) => {
         <AccordionDetails>
           <ItemsStyle isLight={isLight}>
             {category?.subcategories?.map((category) => (
-              <div key={category.name}>{category.name}</div>
+              <div key={category.name}>{capitalizeWord(category.name)}</div>
```
YasielCabrera, line 32: Why not just add the CSS `text-transform: capitalize`? Do we need to add a different behavior?

alejandrocabriales: You are right, I can use this aproach

PR #827, opened 2023-06-12
[SES-268]Create transaction component for tablet devices

YasielCabrera:
# Ticket 

# Description
Created Transaction components for tablet and desktop devices

# What solved
- Should add a table with the incoming and outcoming transactions in the "View transaction history" subsection of the "MakerDAO Funding Overview" section (each cell should be a reusable component) 
- Should add dark mode to the "View transaction history"

### DIFF
--- a/src/stories/containers/TransparencyReport/components/AccountsSnapshot/components/Transaction/Transaction.tsx
+++ b/src/stories/containers/TransparencyReport/components/AccountsSnapshot/components/Transaction/Transaction.tsx
@@ -1,12 +1,60 @@
+import styled from '@emotion/styled';
 import { useMediaQuery } from '@mui/material';
 import lightTheme from '@ses/styles/theme/light';
 import React from 'react';
 import MobileTransaction from './MobileTransaction';
+import TransactionAmount from './segments/TransactionAmount';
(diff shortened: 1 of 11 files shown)


src/stories/containers/TransparencyReport/components/AccountsSnapshot/components/Transaction/Transaction.tsx
```diff
@@ -1,12 +1,60 @@
...
+  txHash,
+  counterPartyName,
+  counterPartyAddress,
+  amount,
+  isIncomingTransaction = true,
+}) => {
   const isMobile = useMediaQuery(lightTheme.breakpoints.down('table_834'));
 
-  return isMobile ? <MobileTransaction /> : <div>Transaction</div>;
+  return isMobile ? (
```
alejandrocabriales, line 36: A question here.it's possible to have this approach with CSS?

YasielCabrera: Yes, it is possible, but in this case, as those two components are very complex and have a lot of events and behavior I wanted to remove them entirely from the DOM instead of just hiding it to avoid performance issues or unexpected behavior

PR #842, opened 2023-06-16
[SES-268] Integrate on chain reserve cards with the API

YasielCabrera:
# Ticket

# Description
Render the On Chain reserves cards according to the data coming from the API

# What solved
- Should render the "On chain reserves" cards based on the data coming from the API (each card with the correct data)


alejandrocabriales (approved):
LGTM

PR #846, opened 2023-06-20
[No ticket-278] Add mark in case the value do not exist

alejandrocabriales:
# Ticket

# What solved
✅The view should be properly loaded.


YasielCabrera (approved):
LGTM

PR #848, opened 2023-06-20
[No-ticket-278] Add default values in case new category is add

alejandrocabriales:
# Ticket

# What solved
✅The view displays an error.

# Description
Add default color when a new category is added
Fix the problem in Activity Global

### DIFF
--- a/src/stories/components/CategoryChip/CategoryChip.tsx
+++ b/src/stories/components/CategoryChip/CategoryChip.tsx
@@ -59,8 +59,19 @@ const colors: { [id: string]: any } = {
     colorDark: '#34AAFF',
     backgroundDark: '#4992ff26',
   },
-  // TODO: Add correct color when it's ready
   Legal: {
+    color: '#5D48FF',
+    background: '#F7F5FF',
+    colorDark: '#5D48FF',
+    backgroundDark: '#54697826',
+  },
+  ScopeFacilitator: {
+    color: '#FF78F2',
+    background: '#F7F5FF',
+    colorDark: '#FF78F2',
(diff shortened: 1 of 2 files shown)


src/stories/components/CategoryChip/CategoryChip.tsx
```diff
@@ -70,6 +81,19 @@ const colors: { [id: string]: any } = {
...
+        color: isLight ? colors.Default.color : colors.Default.colorDark,
+        borderColor: isLight ? colors.Default.color : colors.Default.colorDark,
+        background: isLight ? colors.Default.background : colors.Default.backgroundDark,
+        ...props.style,
+      }}
+    >
+      {props.category}
+    </Chip>;
+  }
+
```
YasielCabrera, line 96: Why not just create a variable or any other method to avoid duplicating the `Chip` code? Maybe something like this:

```tsx
const palette= colors[props.category] ? colors[props.category] : DEFAULT;

<Chip
      style={{
        color: isLight ? palette.color : palette.colorDark,
        borderColor: isLight ?palette.color : palette.colorDark,
        background: isLight ? palette.background : palette.backgroundDark,
        ...props.style,
      }}
    >
      {props.category}
    </Chip>;
```

YasielCabrera (changes requested):

src/stories/components/CategoryChip/CategoryChip.tsx
alejandrocabriales: you are right, let me take this approach

YasielCabrera (approved):
LGTM

PR #849, opened 2023-06-20
[No-ticket-278] Fix breaking change in the global activity feed

alejandrocabriales:
# Ticket

# What solved
✅The data getting from the API should be displayed in the columns defined in the view

# Description
This is to fix the breaking change in the activity feed page

### DIFF
--- a/src/core/models/dto/coreUnitDTO.ts
+++ b/src/core/models/dto/coreUnitDTO.ts
@@ -140,6 +140,11 @@ export interface ActivityFeedDto {
       code: string;
       shortCode: string;
     };
+    owner: {
+      code: string;
+      shortCode: string;
+      type: string;
+    };
     month: string;
     budgetStatementId: number;
   };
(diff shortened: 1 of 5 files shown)


src/core/models/dto/coreUnitDTO.ts
```diff
@@ -140,6 +140,11 @@ export interface ActivityFeedDto {
       code: string;
       shortCode: string;
     };
+    owner: {
+      code: string;
+      shortCode: string;
+      type: string;
+    };
```
YasielCabrera, line 147: Move the params to a different type. Eg:

``` ts
interface CoreUnitParam  {
  coreUnit: {
    code: string;
    shortCode: string;
  }
}

interface OwnerParam {
  owner: {
    code: string;
    shortCode: string;
    type: string;
  }
}

type ActivityFeedParams = CoreUnitParam | OwnerParam
```

alejandrocabriales: You are right

YasielCabrera (approved):
LGTM

PR #856, opened 2023-06-21
[SES-268] Use the same Identicons that Gnosis uses

YasielCabrera:
# Ticket

# Description
Replace the `react-18-blockies` in favor of the same identicon library that Gnosis uses in order to have the same identicon styles per address. Also replaced the identicons used in the CU reports and delegates reports pages

# What solved
- [X] Should make to match the blockies identicon with gnosis (try this [library](https://www.npmjs.com/package/ethereum-blockies-base64) or explore the gnosis source code)


alejandrocabriales (approved):
LGTM

PR #858, opened 2023-06-22
Add Header story for new item

alejandrocabriales:
# Ticket

https://trello.com/c/8LcgvKYi/286-user-story-basic-list-with-ecosystem-actors

# What solved
[X]- Should have the header new link to the header (Ecosystem Actors)

# Description
This is for adding new items and stories to the Header Component


dvegap (approved):

dvegap95 (approved):

dvegap95 (approved):

src/stories/components/Header/Header.tsx
```diff
@@ -64,20 +69,20 @@ const Header: React.FC = () => {
         </ContainerLogoSelect>
 
         <Navigation>
-          {menuItems.map((item: MenuType) => (
-            <Link href={item.link} passHref key={item?.title}>
+          {Object.keys(menuItems).map((item) => (
```
dvegap95, line 72: I'll recommend to iterate the object with both it's key and value so we don't have to acces the value through the key evry time, also I wouldn't use item to reffer to the key of the item but for the value itself, since this way seems counterintuitive, normally it'll go: ```item = object[key]``` instead of ```something=object[item]```

```suggestion
          {Object.entries(menuItems).map(([key, item ]) => (
          // here we can use item instead of menuItems[key] which is cleaner and also provides the target value directly
```

In case you're never going to need the key name, you could even go further and do 
```suggestion
          {Object.values(menuItems).map((item) => (
          // here you can do item.something with item being the value of that entry on each iteration
```

src/stories/components/TopBarSelect/TopBarSelect.tsx
```diff
@@ -51,15 +51,15 @@ export const TopBarSelect = (props: TopBarSelectProps) => {
               }}
             />
           </CloseWrapper>
-          {menuItems.map((item) => (
-            <Link href={item.link} passHref>
+          {Object.keys(menuItems).map((item) => (
```
dvegap95, line 54: Same as before here

dvegap95 (changes requested):

src/stories/components/Header/Header.tsx
alejandrocabriales: you are right, done

src/stories/components/TopBarSelect/TopBarSelect.tsx
alejandrocabriales: done

YasielCabrera (approved):
LGTM

dvegap95 (approved):

PR #877, opened 2023-07-03
[SES-268] Improve the UI when there is no data for the off-chain section

YasielCabrera:
# Ticket

# Description
Hide the arrow of the CU reserve cards when they have no transactions, also was made it unclickable the cards when there are no transactions. The Off-Chain section and all its data are now hidden when there is no Off-chain data

# What solved
- [X] Should hide the expand/collapse arrow when there are no transactions from the Off-Chain Reserves/On-Chain Reserves
- [X] Should hide the section and the Include Off-Chain Reserves checkbox when there is no data


alejandrocabriales (approved):
LGTM

PR #881, opened 2023-07-05
[SES-268] Add On chain only version to the expense comparison table

YasielCabrera:
# Ticket

# Description
A new version of the table is sown in the Expense comparison section when there are no off-chain data, this change includes the mobile version of the table in cards too. The performance of the Snapshot account page was improved by memorizing the high CPU-demanding operations. 

# What solved
- [X] Should show a compacted version of the expense comparison section when there are no off-chain data


alejandrocabriales (approved):
LGTM

PR #884, opened 2023-07-05
Add new page about to the route

alejandrocabriales:
# Ticket

# What solved
- [X]  Should create a new route for the profile page with feature flags
- [X] Add scaffold for the page
- [X]  Should add a link to navigate to this page the in the actor table

# Description
This adds a scaffold for the page and adds feature flags


feature-flags/feature-flags.production.ts
```diff
@@ -14,4 +14,5 @@ export const flagsProduction: FeatureFlagsInterface = {
   FEATURE_ACCOUNTS_SNAPSHOT: false,
   FEATURE_TEMPORARY_ACCOUNTS_SNAPSHOT_PAGE: false,
   FEATURE_ECOSYSTEM_ACTORS: false,
+  FEATURE_ECOSYSTEM_ACTORS_ABOUT: true,
```
YasielCabrera, line 17: This page is under development so this should be set to false on production

feature-flags/feature-flags.interface.ts
```diff
@@ -10,6 +10,7 @@ export interface FeatureFlagsInterface {
   FEATURE_TRANSPARENCY_COMMENTS: boolean; // disable Comments Tab
   FEATURE_RECOGNIZED_DELEGATES: boolean; // disable  Recognized Delegates page
   FEATURE_ACCOUNTS_SNAPSHOT: boolean; // disable Accounts Snapshot tab in the expense reports tab
-  FEATURE_ECOSYSTEM_ACTORS: boolean; // disable Accounts Snapshot page
+  FEATURE_ECOSYSTEM_ACTORS: boolean; // disable Actors Snapshot page
```
YasielCabrera, line 13: Could you update the comment as this is not a "snapshot page"

feature-flags/feature-flags.interface.ts
```diff
@@ -10,6 +10,7 @@ export interface FeatureFlagsInterface {
...
   FEATURE_TEMPORARY_ACCOUNTS_SNAPSHOT_PAGE: boolean; // disable Accounts Snapshot page
+  FEATURE_ECOSYSTEM_ACTORS_ABOUT: boolean; // disable Actors Snapshot page
```
YasielCabrera, line 15: Could you update the comment as this is not a "snapshot page"

alejandrocabriales: you are right

feature-flags/feature-flags.interface.ts
alejandrocabriales: feature/ecosystem-actor-profile-page

feature-flags/feature-flags.production.ts
alejandrocabriales: done

PR #885, opened 2023-07-05
[SES-268] Enabled the account snapshots tab

YasielCabrera:
# Ticket 

# Description
Activated the Accounts snapshots tab in the CU reporting page with the general data coming from the API

# What solved
- [X] Should move the accounts snapshot as a tab in the CU reporting page (just before the comments tab)


alejandrocabriales (approved):
LGTM

PR #888, opened 2023-07-07
[SES-268] Integrate the expense comparison section with the API

YasielCabrera:
# Ticket

# Description
Integrate the Expense comparison section with the API
Renamed the new Tooltip component

# What solved
- [X] Should display the correct data coming from the API in the Reported Expenses Comparison table/cards
- [X] Should hide the totals row if there is only one month in reported comparison table


alejandrocabriales (approved):
LGTM

PR #891, opened 2023-07-11
Add Header Navigation

alejandrocabriales:
# Ticket

# What solved
- [X]  Should create show the correct data navigation subheader component

# Description
This adds header navigation stories and put in a component that can be used in all app


YasielCabrera (approved):
LGTM

PR #895, opened 2023-07-12
[SES-268] Update and cache the account Snapshots on each month change

YasielCabrera:
# Ticket

# Description
Each time the user changes the month in the CU reporting page and the account snapshots tab is activated, the data changes and is cached to avoid re-fetching the same data.

# What solved
- [X] Should update the data every time a budget statement month is changed in the CU reporting page in order to show the account snapshot data for the selected month
- [X] Should cache the state in order to avoid re-fetching data that was already fetched


alejandrocabriales (approved):
LGTM

PR #898, opened 2023-07-13
Add logic for the render actor page

alejandrocabriales:
# Ticket

# What solved
- [X]  Should create a component for makerdao for the new description and sentence the description
- [X]  Should have  A "Services Offered" section listing services the actor provides, each with a brief (1-2 sentences) description.

# Description
Change the text Core Unit table and adjust the table
Add the link to each month

### DIFF
--- a/src/core/context/ActorContext.tsx
+++ b/src/core/context/ActorContext.tsx
@@ -1,0 +1,16 @@
+import { useContext, createContext } from 'react';
+import type { EcosystemActor } from '../models/dto/teamsDTO';
+
+export type CoreUnitContextValues = {
+  actors?: EcosystemActor[];
+  actor?: EcosystemActor;
+  setCurrentActor: (actor: EcosystemActor) => void;
+};
+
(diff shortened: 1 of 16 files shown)


src/core/context/ActorContext.tsx
```diff
@@ -0,0 +1,16 @@
+import { useContext, createContext } from 'react';
+import type { EcosystemActor } from '../models/dto/teamsDTO';
+
+export type CoreUnitContextValues = {
```
YasielCabrera, line 4: Note that this is an `AuditorContext` so the name should be according to that and avoid to use CoreUnit here

YasielCabrera (changes requested):

src/core/context/ActorContext.tsx
alejandrocabriales: you are right, done

YasielCabrera (approved):
LGTM

PR #918, opened 2023-07-19
[SES-288] Fix shortcode issues

YasielCabrera:
# Ticket
https://trello.com/c/KT93qSRs/288-user-story-ability-to-view-ecosystem-actor-expense-reports

# Description
Fix the issues

# What solved
- [X] When there is no data to show, a message should be displayed including the ecosystem actor shortcode.
- [X] The "Visit makerburn.com view the... " text should have the same shortcode in all tabs of this view.  **Current Output:** The Auditor view takes a different value than the rest of the tabs.
- [X] When there is no data to show, a message should be displayed referring to Ecosystem Actor after the shortcode.

### DIFF
--- a/src/stories/components/AdvancedInnerTable/AdvancedInnerTable.tsx
+++ b/src/stories/components/AdvancedInnerTable/AdvancedInnerTable.tsx
@@ -247,7 +247,7 @@ export const AdvancedInnerTable: React.FC<AdvancedInnerTableProps> = ({
       </CardsWrapper>
     </>
   ) : (
-    tablePlaceholder ?? <TransparencyEmptyTable longCode={longCode} />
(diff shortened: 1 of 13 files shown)


src/stories/components/AdvancedInnerTable/AdvancedInnerTable.tsx
```diff
@@ -247,7 +247,7 @@ export const AdvancedInnerTable: React.FC<AdvancedInnerTableProps> = ({
       </CardsWrapper>
     </>
   ) : (
-    tablePlaceholder ?? <TransparencyEmptyTable longCode={longCode} />
+    tablePlaceholder ?? <TransparencyEmptyTable longCode={longCode} shortCode={longCode} />
```
alejandrocabriales, line 250: Is this correct because you are using longCode for shortCode?

PR #921, opened 2023-07-20
Feature add filter actor page

alejandrocabriales:
# Ticket

https://trello.com/c/62d6Vin8/303-user-story-basic-list-with-ecosystem-actors-filtering

# What solved
- [X] Should Clicking on an actor in the list must redirect users to a detailed profile page for that actor
- [X] Should the filter show the correct result

# Description
This should add a filter to the page


YasielCabrera (approved):
LGTM

PR #923, opened 2023-07-24
Add image and correct text when not filter match

alejandrocabriales:
# Ticket

# What solved
- [X]  If there are no results, a message should be displayed. **Current Output:** The list is displayed empty. 

# Description
This fixes the error when there are no filters that match


YasielCabrera (approved):
LGTM

PR #928, opened 2023-07-25
Add correct number filters and new colors

alejandrocabriales:
# Ticket

# What solved
- [X] Show only the following categories in the EA list filter: All Active Ecosystem, Actor, Scope Facilitator, Advisory Council Member. if you could add the spaces in between the words too would be great

# Description
Add new required in the filter


YasielCabrera (approved):
lgtm

PR #929, opened 2023-07-25
[SES-268] Expense reporting logic improvements

YasielCabrera:
# Ticket

# Description
A few improvements in the Account snapshot tab were made. The whole logic of the expense reporting pages is now fully reutilizable through hooks

# What solved
- [X] Remove the "via DSSBlow" text from the Funds Returned card.
- [X] The Snapshot Accounts should be displayed as the first tab and selected by default. 
- [X] The "since month" should be bolded.


alejandrocabriales (approved):
LGTM

PR #931, opened 2023-07-31
[SES-QC-278] Actuals breakdown for non headcount expenses without value

YasielCabrera:
# Ticket

# Description
The issue was when the headcountExpense is undefined instead of a boolean, now it is handles as false in those cases

# What solved
- [X]  The response in the API suggests that maybe the canonical categories that are blank are being hidden but the values included in the total. For example 'feeds' and 'marketing costs'. [image.png](https://trello.com/1/cards/647070a1163c44350d910535/attachments/64c75cf8805cc6ffabf58362/download/image.png)


alejandrocabriales (approved):
LGTM

PR #935, opened 2023-08-02
Fix/global activity feed

alejandrocabriales:
# Ticket

https://trello.com/c/HoQPLBoS/289-user-story-view-expense-report-updates-in-activity-feed-last-modified-column-of-ecosystem-actor-list

# What solved
- [X]  The Time is not displayed therefore the rest of the data is displayed in the wrong columns.
- [X]  The view should direct to the Ecosystem Actors Expense Reports view and the menu should indicate the Ecosystem Actors item. 

# Description
This shows the activity feed for the actors, delegate, and core units


YasielCabrera (approved):
lgtm

PR #937, opened 2023-08-03
[SES-268] Fix tooltip resource type

YasielCabrera:
# Ticket

# Description
Show the correct resource name in the snapshot tooltips

# What solved
- [X] Should show the correct resource type instead of a hardcoded "Core Unit" label. Eg: if we're on the Ecosystem actor's page, it should say Ecosystem Actor instead**Current Output:** Core Units is always visible.


alejandrocabriales (approved):
LGTM

PR #947, opened 2023-08-09
[SES-278] Fix tooltip too close to the viewport

YasielCabrera:
# Ticket 

# Description
Prevent the tooltip to be too close to the edge of the window

# What solved
- [X] The tooltip is closer to the right edge.


alejandrocabriales (approved):
LGTM

PR #977, opened 2023-08-29
[SES-292] Update the navigation tabs on scroll

YasielCabrera:
# Ticket
https://trello.com/c/ERuAdkUG/292-user-story-detailed-insight-into-endgame-budget-structure

# Description
Update the URL hash and the navigation tabs on scroll on the endgame page

# What solved
- [X] Should update the navigation tab when scrolling and a section is visible to the user


alejandrocabriales (approved):
LGTM

PR #986, opened 2023-09-01
[SES-293] add card & toggle button group

YasielCabrera:
# Ticket
https://trello.com/c/K6BCBa9u/293-user-story-visualization-of-budget-transition-status

# Description
Added the card for the Transition status section of the endgame page + the button group (without behavior)

# What solved
- [X] Should add a card to contain the chart
- [X] Should add a toggle button group to select between the Actuals and the Budget cap


alejandrocabriales (approved):
LGTM

PR #990, opened 2023-09-04
Add Items for the Header

alejandrocabriales:
# Ticket

# What solved
- [X]  Add each item of the header ( Contributors , Reporting Month, Etc.)

# Description
Add the header for the table in expense


src/stories/containers/Finances/components/DelegateExpenseTrend/HeaderDelegateExpense.stories.tsx
```diff
@@ -0,0 +1,249 @@
...
+            marginLeft: 22,
+          },
+          [lightTheme.breakpoints.up('desktop_1440')]: {
+            marginLeft: 92,
+            justifyContent: 'center',
+          },
+        },
+        sortReverse: true,
+      },
+    ],
```
YasielCabrera, line 94: would be good to export this in the component and use it here, if something of this changes here chromatic will pass but the actual component is going to have an issue that is not going to be detected, and if it is changed in the component but not here, then we're going to have an issue here at chromatic

src/stories/containers/Finances/components/DelegateExpenseTrend/HeaderDelegateExpense.stories.tsx
```diff
@@ -0,0 +1,249 @@
...
+import lightTheme from '@ses/styles/theme/light';
+import HeaderDelegateExpense from './HeaderDelegateExpense';
+import type { ComponentMeta } from '@storybook/react';
+
+export default {
+  title: 'Components/NewFinances/HeaderDelegateExpense',
+  component: HeaderDelegateExpense,
+  parameters: {
+    chromatic: {
+      viewports: [1194, 1280, 1440, 1920],
```
YasielCabrera, line 11: you should pass this parameter per story as the different stories handle different viewports. eg: `HeaderDelegate1194` only works with 1194 breakpoint

YasielCabrera (changes requested):

src/stories/containers/Finances/components/DelegateExpenseTrend/HeaderDelegateExpense.stories.tsx
alejandrocabriales: done

src/stories/containers/Finances/components/DelegateExpenseTrend/HeaderDelegateExpense.stories.tsx
alejandrocabriales: done

YasielCabrera (approved):
lgtm

PR #992, opened 2023-09-04
[SES-QC] Fix restoring expense report content

YasielCabrera:
# Ticket

# Description
By default restoring the expense report content was showing the "Expense report" content in auditor view instead of "Account Snapshots"

# What solved
- [X] When clicking the Expense Reports, the expense reports page opens with the Snapshot Report tab activated, but the content shows the Expense Report, intended behavior is that it shows the corresponding Snapshot Report content.


alejandrocabriales (approved):
LGTM

PR #997, opened 2023-09-06
fix: apple devies endgame scroll

YasielCabrera:
# Ticket
https://trello.com/c/ERuAdkUG/292-user-story-detailed-insight-into-endgame-budget-structure

# Description
Solved issue on iphone devices in the endgame navigation component

# What solved
- [X] **Expected Output:** The section should be displayed at the top of the screen the tab should be marked. **Current Output:** The section is displayed at the top of the screen but the Key Changes tab is still marked (scrolling down has the same behavior).

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
     "graphql": "^16.5.0",
     "graphql-request": "^4.3.0",
     "identicon.js": "^2.3.3",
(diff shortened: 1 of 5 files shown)


alejandrocabriales (approved):
LGTM

PR #1005, opened 2023-09-11
[No-ticket-278]Add forwardRef for the actor summary

alejandrocabriales:
# Ticket

# What solved
- [X]  Core Units view. Scroll down slightly on the information until the header just closes and it will bounce up and down given a specific mouse position.  There is a point on the Core Unit info page where the header section will bounce between hidden and showing.

# Description
Fix the issue when user scroll on the header


YasielCabrera (approved):
lgtm

PR #1016, opened 2023-09-13
Add fix header in CoreUnitSummary

alejandrocabriales:
# Ticket

https://trello.com/c/udGDe7fW/278-qc-round-1-r

# What solved
- [X]  Core Units view. Scroll down slightly on the information until the header just closes and it will bounce up and down given a specific mouse position. **Current Output:** There is a point on the Core Unit info page where the header section will bounce between hidden and showing. (Transparency Report and About Page of Core Units)

# Description
Add fix in the header summary for the page of about page of core unit and transparency report for core unit


YasielCabrera (approved):
lgtm

PR #1036, opened 2023-09-28
Add header for quarterly

alejandrocabriales:
# Ticket

# What solved
- [X] Should has a header quarterly if the periods is select with the respective metric

# Description
Add header for quarterly when the period is selected

### DIFF
--- a/src/stories/containers/Finances/components/ConditionalWrapper/ConditionalWrapper.tsx
+++ b/src/stories/containers/Finances/components/ConditionalWrapper/ConditionalWrapper.tsx
@@ -1,0 +1,41 @@
+import styled from '@emotion/styled';
+import Container from '@ses/components/Container/Container';
+import lightTheme from '@ses/styles/theme/light';
+import React from 'react';
+
+interface Props extends React.PropsWithChildren {
+  period: string;
+}
+
+const ConditionalWrapper: React.FC<Props> = ({ children, period }) => {
+  const Wrapper =
+    period === 'Quarterly' ? <ContainerTable>{children}</ContainerTable> : <Container>{children}</Container>;
+  return Wrapper;
+};
+export default ConditionalWrapper;
+const ContainerTable = styled.div({
+  paddingLeft: 16,
+  paddingRight: 16,
+  [lightTheme.breakpoints.up('tablet_768')]: {
+    paddingLeft: 32,
+    paddingRight: 32,
+  },
+
+  [lightTheme.breakpoints.up('desktop_1280')]: {
+    marginLeft: 'auto',
+    marginRight: 'auto',
+    paddingLeft: 0,
+    paddingRight: 0,
+    maxWidth: 1184,
+  },
+
+  [lightTheme.breakpoints.up('desktop_1440')]: {
+    maxWidth: 1312,
+  },
+  [lightTheme.breakpoints.up('desktop_1920')]: {
+    maxWidth: 'revert',
+    marginLeft: 64,
+    marginRight: 64,
+    justifyContent: 'center',
+  },
+});
--- a/src/stories/containers/Finances/components/ConditionalWrapper/ConditionalWrapper.tsx
+++ b/src/stories/containers/Finances/components/ConditionalWrapper/ConditionalWrapper.tsx
@@ -1,17 +1,10 @@
 import styled from '@emotion/styled';
-import Container from '@ses/components/Container/Container';
 import lightTheme from '@ses/styles/theme/light';
 import React from 'react';
 
-interface Props extends React.PropsWithChildren {
-  period: string;
-}
-
-const ConditionalWrapper: React.FC<Props> = ({ children, period }) => {
-  const Wrapper =
-    period === 'Quarterly' ? <ContainerTable>{children}</ContainerTable> : <Container>{children}</Container>;
-  return Wrapper;
-};
+const ConditionalWrapper: React.FC<React.PropsWithChildren> = ({ children }) => (
+  <ContainerTable>{children}</ContainerTable>
+);
 export default ConditionalWrapper;
 const ContainerTable = styled.div({
   paddingLeft: 16,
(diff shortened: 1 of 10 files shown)


src/stories/containers/Finances/components/ConditionalWrapper/ConditionalWrapper.tsx
```diff
@@ -0,0 +1,41 @@
+import styled from '@emotion/styled';
+import Container from '@ses/components/Container/Container';
+import lightTheme from '@ses/styles/theme/light';
+import React from 'react';
+
+interface Props extends React.PropsWithChildren {
+  period: string;
```
YasielCabrera, line 7: We should use an enum or a `type` for the periods ('Anually', 'Quarterly', etc) in order to avoid error with the values

src/stories/containers/Finances/components/ConditionalWrapper/ConditionalWrapper.tsx
```diff
@@ -0,0 +1,41 @@
...
+}
+
+const ConditionalWrapper: React.FC<Props> = ({ children, period }) => {
+  const Wrapper =
+    period === 'Quarterly' ? <ContainerTable>{children}</ContainerTable> : <Container>{children}</Container>;
+  return Wrapper;
```
YasielCabrera, line 13: you could return it directly to avoid unnecessary variables (which should be in lower case btw due name conventions)

src/stories/containers/Finances/components/HeaderTable/HeaderQuarterly/CellQuarterly.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+  return (
+    <MainContainer isLight={isLight} isTotal={isTotal} className={className}>
+      <ContainerCell>
+        <Quarterly isLight={isLight}>{quarterly}</Quarterly>
+        <ContainerMetricsData>
+          {metrics?.map((metric, index) => (
+            <Metrics key={index}>
+              <Name isLight={isLight}>{returnShortNameForMetric(metric).name}</Name>
+              <Amount isLight={isLight}>{metric.amount.toLocaleString('es-US')}</Amount>
```
YasielCabrera, line 27: We have a helper to avoid using `toLocaleString` directly so we manage the decimal places too: `usLocalizedNumber`

src/stories/containers/Finances/components/HeaderTable/HeaderQuarterly/CellQuarterly.tsx
```diff
@@ -0,0 +1,152 @@
...
+import { returnShortNameForMetric } from '@ses/containers/Finances/utils/utils';
+import { useThemeContext } from '@ses/core/context/ThemeContext';
+import lightTheme from '@ses/styles/theme/light';
+import React from 'react';
+import type { MetricsWithAmount } from '@ses/containers/Finances/utils/types';
+import type { WithIsLight } from '@ses/core/utils/typesHelpers';
+
+interface Props {
+  metrics: MetricsWithAmount[];
+  quarterly: string;
```
YasielCabrera, line 11: Would be a good idea to use the `enum` I propose in my previous comment here too

src/stories/containers/Finances/utils/types.ts
```diff
@@ -41,3 +41,8 @@ export interface MomentDataItem extends Team {
   totalActuals: number;
   lastModified: DateTime;
 }
+export type Metrics = 'Budget' | 'Actual' | 'Forecast' | 'Net Expenses On-chain' | 'Net Expenses Off-chain';
```
YasielCabrera, line 44: This should be in singular as it is all the possible metrics but when it is used it represent only one metric unless it is used as an array

src/stories/containers/Finances/utils/types.ts
```diff
@@ -41,3 +41,8 @@ export interface MomentDataItem extends Team {
...
+export type MetricsWithAmount = {
```
YasielCabrera, line 45: use `interface` for consistency as it is an object

YasielCabrera (changes requested):

src/stories/containers/Finances/components/ConditionalWrapper/ConditionalWrapper.tsx
alejandrocabriales: you are right, done

src/stories/containers/Finances/components/HeaderTable/HeaderQuarterly/CellQuarterly.tsx
alejandrocabriales: you are right, done

src/stories/containers/Finances/components/HeaderTable/HeaderQuarterly/CellQuarterly.tsx
alejandrocabriales: For this , quarterly , its indeed an string, its not ('Annually', 'Quarterly', etc.), instead its the string with the form of (Q1-2023,Q2-2023,Q3-2023,Q4-2023)

src/stories/containers/Finances/components/ConditionalWrapper/ConditionalWrapper.tsx
alejandrocabriales: you are right, this is something that this functionality is show betters when add other header, than will made sense the condition
Done

src/stories/containers/Finances/utils/types.ts
alejandrocabriales: you are right, done

src/stories/containers/Finances/utils/types.ts
alejandrocabriales: you are right, done

YasielCabrera (approved):
lgtm

PR #1038, opened 2023-09-29
Feature/header table expense annually monthly

alejandrocabriales:
# Ticket

# What solved
- [X] Should has a header annual if the periods is select with the respective metric
- [X]  Should has a header monthly if the periods is select with the respective metric

# Description
Add other version of the table header monthly and annually


YasielCabrera (approved):
lgtm

PR #1040, opened 2023-10-02
Add mock for the style of table

alejandrocabriales:
# Ticket

# What solved
- [X]  Should add a mock for the table with the quarterly select period is select

# Description
Add styles of the table with mock data, when quarterly filter is select


src/stories/containers/Finances/components/HeaderTable/FinancesTable.tsx
```diff
@@ -0,0 +1,131 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
```
YasielCabrera, line 1: Please, avoid to disable globally eslint, if it is needed to disable a var due this is WIP, do it online, so we remember to delete this after the implementation is complete

YasielCabrera (changes requested):

src/stories/containers/Finances/components/HeaderTable/FinancesTable.tsx
alejandrocabriales: you are right , done

YasielCabrera (approved):
lgtm

PR #1041, opened 2023-10-03
Add style total row and fix number according dimensions

alejandrocabriales:
# Ticket

# What solved
- [X] Should have each main row a total. The color of this cell should be the same color displayed on the Total header.  
- [X] Should have each row a value related to the metrics selected.

# Description
Add different cells values with the different breakpoints


YasielCabrera (approved):
lgtm

PR #1048, opened 2023-10-05
fix: hover on mobile devices in the social icons

YasielCabrera:
# Ticket

# Description
Fixed the issue in the CU, EA and RD summary component

# What solved
- [X] Ecosystem Actors profile and CU About views. Social media icons. Selecting an icon. **Expected Output:** The expected tab should be open and the icon should remain with the green color. **Current Output:** The icon is displayed with black color (hover state).


alejandrocabriales (approved):
LGTM

PR #1051, opened 2023-10-10
[SES-294] Fix finances filters dropdown

YasielCabrera:
# Ticket

# Description
Fix some issues related to the filter select components

# What solved
- [X] The area with options should stay in the same position always, even when you start scrolling the page. **Current Output:** After clicking on period filters, the area with options that appear is moving when scrolling the page.
- [X]  The area with the options for period filter should be aligned to the right side, the same as for metrics filters. **Current Output:** The area that appears after clicking on the period filters is not aligned as the metrics filters are. 
- [X] The window with options should be displayed fully on the screen. **Current Output:** The window seems to be cut.


alejandrocabriales (approved):
LGTM

PR #1056, opened 2023-10-11
[SES-310] Fix delegates NaN values

YasielCabrera:
# Ticket

# Description
Fix the progress bars on the delegates page that were showing no progress due to NaN values

# What solved
- [X] he view should display the data coming from the API.  **Current Output: ** From the API is coming data but the progress bar doesn't show any progress.


alejandrocabriales (approved):
LGTM

PR #1057, opened 2023-10-11
[SES-310] Fix enabling comment tab on auditor view on nav

YasielCabrera:
# Ticket

# Description
Fix the comment tab on the auditor view when the user navigates to the view and the auditor view is the preferred

# What solved
- [X] From the Expense Reports page, Auditor view Comments tab. Going back to the Core Units (selecting IS) and going through the activity feed option. Selecting an activity.  **Expected Output:** Should direct to the comments tab and display the proper data. **Current Output: **It directs to the Comments tab but displays the Account Snapshot data.


alejandrocabriales (approved):
LGTM

PR #1058, opened 2023-10-11
Adjust data when there is more one metric

alejandrocabriales:
# Ticket

# What solved
- [X] Should adjust the data on the row, depending on the metric selected in the Quarterly period.
- [X] Should present the tables ordered by the one with the most sub-categories.    

# Description
Adjust the row when there is more than one metric
Order the table in function number of category


YasielCabrera (approved):
lgtm

PR #1062, opened 2023-10-16
Add others rows if there are more than 16 rows amount all tables

alejandrocabriales:
# Ticket

# What solved
- [X] Should have the table show (Others) rows, when there is more subcategory to show, and this should be a resume of the rest.

# Description
Add others row when there is more than 16 rows, and only show 16 rows as max numbers of row


src/stories/containers/Finances/components/FinacesTable/FinancesTable.tsx
```diff
@@ -29,17 +30,19 @@ const FinancesTable: React.FC<Props> = ({ className, breakdownTable, metrics, pe
   const showQuarterly = !isMobile && period === 'Quarterly';
   const showMonthly = desk1440 && period === 'Monthly';
   const arrayMetrics = new Array<number>(iteration).fill(0);
-  const showFooter = true;
-  // Show color for others depending if odd or even
-  const isPair = orderData[`${tables[tables.length - 1]}`]?.length % 2 === 0;
+
+  // Show color for others depending if odd or pair
+  const isPair = showFooterAndCorrectNumber[`${tables[tables.length - 1]}`]?.length % 2 === 0;
```
YasielCabrera, line 35: please, note here that it should be "isEven", in English we say "odd or even", not "odd or pair"

src/stories/containers/Finances/utils/utils.ts
```diff
@@ -716,3 +717,51 @@ export const sortDataByElementCount = (data: MockData) => {
 
   return sortedData;
 };
+
+export const showOthersFooterRow = (data: MockData) => {
+  let totalRows = 0;
+  Object.keys(data).forEach((key) => {
+    totalRows += data[key].length;
+  });
+  console.log({ totalRows });
```
YasielCabrera, line 726: did you forgot to delete this ` console.log` ?

src/stories/containers/Finances/utils/utils.ts
```diff
@@ -716,3 +717,51 @@ export const sortDataByElementCount = (data: MockData) => {
...
+  const firstElements: Record<string, QuarterlyBudget[]> = {};
+  const totalRows = Object.keys(data).reduce((acc, key) => acc + data[key].length, 0);
+  // Si la cantidad total de filas es menor o igual a 16, no necesitas hacer cambios
+  if (totalRows < maxRows) {
+    return data;
+  }
+
+  // Save the first element of each table
+  Object.keys(data).forEach((key) => {
+    firstElements[key] = [data[key][0]];
```
YasielCabrera, line 754: are we 100% sure that `data[key]` will always be an array with at least 1 element? If not, we should prepare to receive empty arrays here, if that happens we could have an error here

YasielCabrera (changes requested):

src/stories/containers/Finances/utils/utils.ts
alejandrocabriales: You are right

src/stories/containers/Finances/utils/utils.ts
alejandrocabriales: yes, done

src/stories/containers/Finances/components/FinacesTable/FinancesTable.tsx
alejandrocabriales: you are right, done

YasielCabrera (approved):
lgtm

PR #1068, opened 2023-10-20
fix: budget statement navigation

YasielCabrera:
# Ticket

# Description
Allow to navigate the budget statements having in count if there is snapshot data or not

# What solved
- [X] SES core unit. Expense Reports view. Month navigation component.  **Expected Output:** It should allow navigation to the current month.  **Current Output: ** It doesn't allow navigation to the October month even when there is data to show.


alejandrocabriales (approved):
LGTM

PR #1070, opened 2023-10-20
[SES-310] Add delegates auditor flow

YasielCabrera:
# Ticket

# Description
Enable the auditor flow in the delegates reports

# What solved
- [X] If the user logged is an Admin and the same time Auditor then should be allowed to add new reports and audit reports.


alejandrocabriales (approved):
LGTM

PR #1085, opened 2023-10-27
[SES-299] Show snapshot last update on transparency reporting page

YasielCabrera:
## Ticket
https://trello.com/c/0oAK0jem/299-user-story-tpd-11-projects-overview-v1

## Description
Display the snapshot last update instead of budget statement last update when the account snapshot tab is active and keep the previous behavior when the other tabs are active

## What solved
- [X] Should change the source of the value shown for LAST UPDATE when the user selects the Accounts Snapshot tab. The value should now come from the new ‘created’ field in the ‘Snapshot’ table.
- [X] Should the Last Updated = created datetime when tab with Snapshot Report is selected.
- [X] Should use existing datetime behaviour if Expense Report or Comments are selected.

### DIFF
--- a/src/core/models/dto/snapshotAccountDTO.ts
+++ b/src/core/models/dto/snapshotAccountDTO.ts
@@ -66,6 +66,7 @@ export interface Snapshots {
   end: string | null;
   ownerType: string;
   ownerId: string;
+  created: string | null;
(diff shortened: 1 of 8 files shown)


alejandrocabriales (approved):
LGTM

PR #1096, opened 2023-11-06
[SES-299] Expandable deliverables

YasielCabrera:
## Ticket
https://trello.com/c/0oAK0jem/299-tpd-11-projects-overview-v1

## Description
Improved the deliverables section of the projects

## What solved
- [X] Should display all the details of the deliverables card when the "Detailed" view mode option is selected (description and key results)
- [X] Should allow scrolling when the "detailed" view mode option is selected and there are more than 6 key results on desktop but show all the key results on mobile.
- [X] Should the deliverable have the option to expand/collapse if there are more than 4 key results when the "compacted" view mode detail option is selected


alejandrocabriales (changes requested):

alejandrocabriales (approved):
LGTM

PR #1099, opened 2023-11-08
Remove borders when others is present

alejandrocabriales:
## Ticket
https://trello.com/c/JHKHOrzX/294-bsn-1-budget-summary-navigation

## Description
This remove the borders in the last row when the others rows (footer) appears

## What solved
- [X] FR  MakerDAO Legacy Budgets category. The rows should have straight edges. **Current Output:**  A redounded corner is displayed. 
- [X] FR  The columns should be separated by a line.  The lines between the columns aren't visible.
- [X] FR  The option should be disabled by default due to the filters have the default values. The option is active by default. 
- [X] FR The option should be active allowing resetting the filter values.  Changing the filter values, the Reset option is still disabled not allowing to reset to the default values.


nesdc88 (commented):
LGTM

PR #1104, opened 2023-11-10
[Not-ticket] Revert position absolute

alejandrocabriales:
## Ticket
Fix issues when position revert

## What solved
Fix the position revert in finances page, that made the legend float in wrong position


nesdc88 (commented):
LGTM

PR #1105, opened 2023-11-10
[No-ticket] Fix when there is not name

alejandrocabriales:
## Ticket
Fix the issues when there is not name

## What solved
Default text when there is not name


YasielCabrera (approved):
lgtm

PR #1108, opened 2023-11-14
[EML-1] feat: Should allow to toggle on/off the data throughout the legend on Expense Metrics Line Chart

nesdc88:
## Ticket
https://trello.com/c/Dy3KazCU/305-eml-1-expense-metrics-line-chart

## Description
Allow to toggle on/off the data throughout the legend

## What solved
- [X] **FR7:** Should allow to toggle on and off the data throughout the legend

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -1,11 +1,13 @@
+/* eslint-disable spellcheck/spell-checker */
 /** @type {import('next').NextConfig} */
 const nextConfig = {
   reactStrictMode: true,
+  swcMinify: true,
   compiler: {
     emotion: true,
   },
   images: {
-    domains: ['raw.githubusercontent.com'],
+    unoptimized: true,
   },
 };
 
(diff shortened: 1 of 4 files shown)


next.config.js
```diff
@@ -4,6 +4,9 @@ const nextConfig = {
   compiler: {
     emotion: true,
   },
+  images: {
+    domains: ['raw.githubusercontent.com'],
+  },
```
YasielCabrera, line 9: Images coming from the API could be from dynamic and unknown sources, not necessarily from `githubusercontent.com`, so the in order to prevent crashes would be better to handle the images as `unoptimized`, this way we can still use the `Image` component and render properly any image of any source safely, this next config is not necessary using that approach

YasielCabrera (changes requested):

next.config.js
nesdc88: Got it, check now

YasielCabrera (approved):
lgtm

PR #1115, opened 2023-11-21
[SES-299] Apply filters to supported projects

YasielCabrera:
## Ticket
https://trello.com/c/0oAK0jem/299-tpd-1-team-projects-details-tpd-11-tdp-12-tdp-13-tdp-15

## Description
Apply filter to the supported projects and show "No result founds" if there are no results

## What solved
- [X]  Should apply the filter to the supported projects. If there are no supported projects matching the filters then should show an italic "No results found" text on the supported projects section
- [X] Phoenix Labs ecosystem actor.  Supporter avatar. **Expected Output:** If the project doesn't have any supporters, the avatar and tooltip should not be visible. If there are supporters then the avatar should be displayed properly and the tooltip should list the names. **Current Output:** A little circle is displayed and the tooltip is not displaying any name.


nesdc88 (commented):
LGTM

PR #1118, opened 2023-11-24
[SES-299] Show "no key result" msg on mobile & fix filter chips size

YasielCabrera:
## Ticket
https://trello.com/c/0oAK0jem/299-tpd-1-team-projects-details-tpd-11-tdp-12-tdp-13-tdp-15

## Description
On mobile when the deliverables has no key results it shows a message instead of not displaying anything at all. 

## What solved
- [X] Should the Key Results subtitle be visible even when there are no key results in the deliverable. (Mobile)
- [X] Status filter. **Expected Output:** The status options should be displayed with the size visible in the Core Units view. **Current Output:** The options are displayed bigger than the options visible in the Core Units view


nesdc88 (commented):
LGTM

PR #1128, opened 2023-12-01
[RMV-1] Milestones timeline

YasielCabrera:
## Ticket
https://trello.com/c/HIe66Yme/297-rmv-1-roadmaps-and-milestones-view

## Description
Created the UI for the milestones Roadmap

## What solved
- [X] Should add the milestones cards as a timeline
- [X] Should display the milestones cards above the timeline if there are between 2-4 milestones
- [X] Should show 4 milestones only on the timeline on mobile
- [x] Should add a "View all Roadmap"/"View less Roadmap" button to expand/collapse the milestones on mobile when there are more than 4 milestones

## Screenshots (if apply)
Screenshot_461

### DIFF
--- a/src/stories/containers/EndgamePhaseOneProgress/components/RoadmapTimeline/RoadmapTimeline.tsx
+++ b/src/stories/containers/EndgamePhaseOneProgress/components/RoadmapTimeline/RoadmapTimeline.tsx
@@ -20,7 +20,7 @@ const RoadmapTimeline = () => {
   return (
     <div>
       {isMobile ? (
-        <MobileTimeline isLight={isLight}>
+        <MobileTimeline>
           <MilestoneCard />
           <MilestoneCard />
           <MilestoneCard />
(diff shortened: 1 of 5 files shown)


src/stories/containers/EndgamePhaseOneProgress/components/RoadmapTimeline/RoadmapTimeline.tsx
```diff
@@ -0,0 +1,206 @@
...
+        <MobileTimeline isLight={isLight}>
+          <MilestoneCard />
+          <MilestoneCard />
+          <MilestoneCard />
+          <MilestoneCard />
+          <MilestoneCard />
+
+          <ButtonBox>
+            <Button isLight={isLight}>
+              <span>View all Roadmap</span>
```
alejandrocabriales, line 32: Can you move to style  component

src/stories/containers/EndgamePhaseOneProgress/components/RoadmapTimeline/RoadmapTimeline.tsx
```diff
@@ -0,0 +1,206 @@
...
+              <svg
```
alejandrocabriales, line 33: Its possible to made this a component, also to avoid inline style or you never will use this component again

src/stories/containers/EndgamePhaseOneProgress/components/RoadmapTimeline/RoadmapTimeline.tsx
YasielCabrera: In this case, this is used to create content blocks and not to stylize those blocks

src/stories/containers/EndgamePhaseOneProgress/components/RoadmapTimeline/RoadmapTimeline.tsx
YasielCabrera: This is not going to be reused and it is a really simple svg so there's no need to create a new component. The style in this case is to rotate the SVG and it depends on a state vars, this way is more simple and readable

alejandrocabriales: LGTM

PR #1131, opened 2023-12-04
Feature/doughnut reusable component without swiper

alejandrocabriales:
## Ticket
https://trello.com/c/JHKHOrzX/294-bsn-1-budget-summary-navigation

## Description
Add legend manually and add the same effects as the chart

## What solved
- [X] Should manage legends manually
- [X] Should add the tooltip when hovering in the legend
- [X] Should add the effect of hiding the chart data when clicking
- [X] Should change the format of the legends in the third level so that it is code that is displayed and the items in a single row
- [X] Should hide the filters in the third level


YasielCabrera (approved):
lgtm

PR #1132, opened 2023-12-04
[QC-r2] Fix doughnut chart NaN numbers

YasielCabrera:
## Ticket

## Description
Some numbers were NaN and the chart was not showing up, now the source of the issue is solved

## What solved
-  Recognized Delegate's view. Key Stats section. **Expected Output:**  In the Percentage of Total DAO Expense card if there are no values or the values are zero, the values reflected should be 0 and the pie chart should have a placeholder when there are no values to show. **Current Output: ** NAN and NaN are displayed and the pie chart is not visible causing a big space.


alejandrocabriales (approved):
LGTM

PR #1135, opened 2023-12-06
[QC-r2] accomodate EA scopes

YasielCabrera:
## Ticket

## Description
Now when there are more than 2 scopes they're grouped to fit in the space they have

## What solved
- [X] Ecosystem Actors view. SES EA, Scope column.   **Expected Output:** The scopes should be accommodated in the row when there are more than 3. **Current Output:** Four scores are displayed and two of them are displayed out of the row.

## Screenshots (if apply)
Screenshot_479


alejandrocabriales (approved):
LGTM

PR #1144, opened 2023-12-12
[MER-1] Fix expense reports header

YasielCabrera:
## Ticket

## What solved
- [X] Should be displayed in the table, the column option selected.


alejandrocabriales (approved):
LGTM

PR #1149, opened 2023-12-15
[QC-r2] Fix snapshot expense comparison header

YasielCabrera:
## Ticket

## Description
Fix the empty headers on the snapshot expenses comparison when there are not off-chain data

## What solved
- [X] Expense Reports view (CU and Ecosystem Actors). Account Snapshot tab. Reported Expenses Comparison section. **Expected Output:** When there are no OnChain values, the last column shouldn't be displayed.


alejandrocabriales (approved):
LGTM

PR #1157, opened 2023-12-21
[QC-r2] Fix snapshot expenses with no off chain data

YasielCabrera:
## Ticket

## Description
Fix undefined values on account snapshot

## What solved
- [X] Accounts Snapshot view, Reported Expenses Comparison section. **Expected Output:**   **Current Output: ** REPORTED ACTUALS column is displaying "undefined" text instead of the value. The total percent is displayed with NaN instead of 0.


alejandrocabriales (approved):
LGTM

PR #1189, opened 2024-01-15
[BSN-330]Add limited for the code and change position of percent

alejandrocabriales:
## Ticket

## Description
Solving issues in SB + N

## What solved
- [X]  Number formatting should be adjusted to 3 significant digits + size letter (same as the section on the left 1.49 M). Change the label to percentage number first. **Example: 5.2% (1.49M)** Drop the currency information (replace with information icon that says “All amounts are in DAI”), Legend label characters should be limited to 8 (so for longer words the 8th character can be  “…”).


YasielCabrera (approved):
lgtm

PR #1193, opened 2024-01-17
[QC] Fix total expenses number

YasielCabrera:
## Ticket

## Description
Fix the total expenses number as it does not coincide with the total number of the breakdown table

## What solved
- [X] Homepage (finances). **Expected Output:** The Total Reported Expenses value should be the same as the total of the table. **Current Output: ** Different numbers are displayed in both places.


alejandrocabriales (approved):
LGTM

PR #1202, opened 2024-01-19
[MER-1] Remove budget metric

YasielCabrera:
## Ticket

## Description
Removed the option budget from the metric select in the Expense reports section

## What solved
- [X] Should remove the Budget metric


alejandrocabriales (approved):
LGTM

PR #1205, opened 2024-01-22
[MER-1] Fix expense reports links

YasielCabrera:
## Ticket

## Description
Fix the links on the View button of the expense reports section

## What solved
- [X] Clicking on the View option. **Expected Output:** It should redirect to the specific expense report. **Current Output:** It redirects to the 404 page


alejandrocabriales (approved):
LGTM

PR #1212, opened 2024-01-24
[MER-1] Integrate EMR filters with the API

YasielCabrera:
## Ticket

## Description
Apply sorting and filtering to the expense report items

## What solved
- [X] Should display the proper data coming from the API when the filter is selected.
- [X] Should display the proper data when the status is selected.
- [X] Should ensure the sorting is applied only for Month and Last Modified columns. (Come from the API)


alejandrocabriales (approved):
LGTM

PR #1219, opened 2024-01-24
Issues/sbn version 9

alejandrocabriales:
## Ticket

## Description
Fix issues in the SB+N

## What solved
- [X] Legacy Budgets.  Pie chart, legend, tooltip. - ** **Expected Output:** When there is only one budget with value then the percent should be 100%.  **Current Output:** The percent is displayed less or more than 100%


YasielCabrera (approved):
lgtm

PR #1232, opened 2024-01-27
Fix account snapshot difference percentage

YasielCabrera:
## Ticket
No ticket

## Description
Fix the percentage of the difference in the "Reported Expenses Comparison" section of the account snapshot


alejandrocabriales (approved):
LGTM

PR #1239, opened 2024-01-30
Add limit in the navigations

alejandrocabriales:
## Description
Fix the level in the navigation in the cards

## What solved
- [X] FR - Finances->Scope Frameworks Budget->Protocol Scope->Multichain engineering. The data should match all sections.  The first section and breakdown table do not display any data, however breakdown chart and expense metrics chart display values related to the "Not-name" categories.


YasielCabrera (approved):
lgtm

PR #1240, opened 2024-01-31
[MER-1] Fix expense reports alignment

YasielCabrera:
## Ticket

## Description
Fix expense report issues

## What solved
- [X] Actuals. **Expected Output:** The values should be aligned with the column header **Current Output:** The values are not aligned. 
- [X] Status filter. Dropdown. **Expected Output:** The list should display the number of items for each status. **Current Output:** All statuses are displayed with 0 even when the table shows elements.


alejandrocabriales (approved):
LGTM

PR #1241, opened 2024-01-31
[QC] Fix expense comparison percentage when there are not off chain data

YasielCabrera:
## Ticket

## Description
Show the percentage of the Expense comparison section of the account snapshots correctly when there are not offchain data

## What solved
- [X] SIDESTREAM ecosystem actor. Accounts Snapshot view, Reported Expenses Comparison section. **Expected Output:** The percent should be displayed properly. **Current Output:** When there is no Off- Chain the percent is displayed -0.10% instead of 10%.


alejandrocabriales (approved):
lgtm

PR #1242, opened 2024-01-31
[MER-1] Avoid query revalidation

YasielCabrera:
## Ticket

## Description
Avoid query revalidation on the expense reports page as the data is immutable

## What solved
- [X] Request (paginator).  **Current Output:** Selecting the Load more option, the frontend requests twice the first 10 elements and then requests the second page


alejandrocabriales (approved):
LGTM

PR #1248, opened 2024-02-02
[MER-1] Disable empty expense reports when there is not data

YasielCabrera:
## Ticket

## Description
If there are no expense reports for certain levels the section is hidden

## What solved
- [X] The section should be visible only at the level where there are expense reports available. **Current Output:** The section is displayed without value in several levels


alejandrocabriales (approved):
LGTM

alejandrocabriales (approved):
LGTM

PR #1252, opened 2024-02-04
[BSN-1] Show only one row of navigation cards

YasielCabrera:
## Ticket

## Description
Show 5 navigation cards on desktop and 3 in tablet, if there are more, then show a carrousel to avoid having multiple rows

## What solved
- [X] The navigation cards should be displayed in one row. More than 5, carousels should be introduced.  **Current Output:**  The cards are displayed in two rows and the second one displays a card across the width of the view.


alejandrocabriales (approved):
LGTM

PR #1253, opened 2024-02-05
Fix/issuees chart dont show legend

alejandrocabriales:
## What solved
- [X] Months in the x-axis should be displayed as the design and the rest of the chart shows. **Current Output:** The chart displays numbers instead of months.


YasielCabrera (approved):
lgtm

PR #1265, opened 2024-02-07
[SBN-1] Add correct start point for waterfall

alejandrocabriales:
## What solved
- [X] The values should be calculated this way: 
- (
  JAN.netProtocolOutflow.sum
 - JAN.netProtocolOutflow.value
)
-
(
  JAN.paymentsOnChain.sum
 - JAN.paymentsOnChain.value
)

## Screenshots (if apply)
data-monthly
water-chart-data


YasielCabrera (approved):
lgtm

PR #1268, opened 2024-02-09
[BSN-1] Allow to keep the filter state using url search params in the breakdown table

YasielCabrera:
## Ticket

## Description
When the filters in the breakdown table change the state is saved in the URL, if the URL is opened in other tab/browser the state of the filters is going to be restored.

## What solved
- [X] The view should direct to the Breakdown Table section remaining the filters selected.


alejandrocabriales:
LGTM

PR #1269, opened 2024-02-09
[RC-1] Remove the line at the bottom

alejandrocabriales:
## Ticket
https://trello.com/c/3nvO11Pb/329-fte-1-fte-chart

## Description
Fix lines appear at the bottom

## What solved
- [X] Finances. Year: 2024. Period: Monthly. **Expected Output:** When the value is 0, the values should not be displayed. The x- 
        axis displays the red line and the +0 values in green color. 
- [X] Selecting the Reset option. **Expected Output:** The Reset option should reset the filter to the default state.  The AllMakerDao filter is reset and with that action, the chart displays only the lines.


YasielCabrera (approved):
lgtm

PR #1270, opened 2024-02-09
[BSN-1] Fix first sections charts

YasielCabrera:
## Ticket

## Description
Fix the breakdown chart and the Doughnut chart of the first section when the user browses in deeper levels and goes back to cached data

## What solved
- [X] The view should display all budgets coming from the API. **Current Output:**  The breakdown chart is displaying 2 new values but the rest of the sections don't.


alejandrocabriales (approved):
LGTM

PR #1275, opened 2024-02-12
[BSN-1] Fix resize the windows was redirecting to finances

YasielCabrera:
## Ticket

## Description
When the windows are resized in the finances page it was redirecting to the finances root page when it should keep in the same page

## What solved
- [X] Finances->Scope Frameworks Budget->Support Scope->EAE. Changing the resolution.- ** **Expected Output:**  The current page should remain visible. **Current Output:**  The system resets and the MakerDao level is displayed or the current data is displayed but with the breadcrumb and page title as MakerDao level.


alejandrocabriales (approved):
LGTM

PR #1276, opened 2024-02-12
[RC-1] Fix the problem relate wit the lines in waterfall chart

alejandrocabriales:
## Ticket

## Description
This fix the problem about the lines wrong painting in the bars

## What solved
- [X] Lines connecting the bars. **Expected Output:** The lines should be slim and positioned as defined in the design.  Part of the line is visible on the top/bottom of the bar no as part of it.
- [X]  Finances. Year: 2024. Period: Annually. The values should be reflected from the Start column to the Finish column.  A red line is displayed on the x-axis after the Finish column.


YasielCabrera (approved):
lgtm

PR #1281, opened 2024-02-13
[RC-Not-ticket] Fix color Linea in the waterfall chart

alejandrocabriales:
## Ticket
No-ticket

## Description
Fix the correct color from each bar 

## What solved
- [X] Add correct color of the Linea in the chart


YasielCabrera (approved):
lgtm

PR #1284, opened 2024-02-14
Fix the filters per level

alejandrocabriales:
## Ticket

## Description
Show the correct filers in each level

## What solved
- [X]  Finance->MakerDAO Legacy Budget->core-units. AllMakerDao filter, default state. **Expected Output:** All elements should be selected and the filter should indicate the amount of the elements selected.  **Current Output:** A wrong number is displayed and the elements of the list are not selected.


YasielCabrera (approved):
lgtm

PR #1287, opened 2024-02-14
Fix the color line in dark mode

alejandrocabriales:
## Ticket

## Description
Fix the colors of the line in dark mode and when not change values put green color for the lines

## What solved
- [X]  Change the red color to #641E08.
- [X] Lines. When there is no change in the values, show the lines in green.


YasielCabrera (approved):
lgtm

PR #1297, opened 2024-02-17
[BSN-1] Fix on chain metric on the expense reports section

YasielCabrera:
## Ticket

## Description
Remove off chain metric and set the on chain metric in the expense reports filter

## What solved
- [X]  **- Expense Reports section, Metrics filter.- ** **Expected Output:** Net Expenses On-chain should be listed in the metrics. **Current Output:**  The metric Net Expenses On-chain is missing and instead appears as Net Expenses Off-chain.


alejandrocabriales (approved):
LGTM

PR #1299, opened 2024-02-19
[BSN-1] Restore max amount of metric in annual when the granularity is changed to Annual

YasielCabrera:
## Ticket

## Description
If the period was Monthly or Quarterly and then change to Annually the selected metrics was just 1-3 but they had to be the max amount of allowed metrics for the annually period

## What solved
- [X] Finances->Scope Frameworks Budget. Breakdown table, Changing the period filter. -** **Expected Output:** The metrics should change to the default state depending on the period selected. ** Current Output:** After selecting the Monthly period, the metrics filter remains only one metric selected in all the periods selected.


alejandrocabriales (approved):
LGTM

PR #1311, opened 2024-02-22
fix: reset filter hability

YasielCabrera:
## Ticket

## Description
The reset filter button wasn't disabled sometimes when the default filters were selected

## What solved
- [X]  Finances view. Breakdown table section. Reset filters option. -** **Expected Output:** The reset filters option should be disabled by default, It should be active only when the filters are changed. **Current Output:** The option always has the active state.


alejandrocabriales (approved):
LGTM

PR #1312, opened 2024-02-22
[RC-1] Included all the analytics for the level budgets

alejandrocabriales:
## Ticket

## Description
Fix the analytics for the waterfall

## What solved
- [X]  The difference between the Protocol Outflow and Payment On-chain line charts needs to be the same as the reserves increment/decrement for that budget level. Nov 2022 bar for atlas/legacy/core-units should show +3.41M instead of -655K. .


YasielCabrera (approved):
lgtm

PR #1314, opened 2024-02-22
[BSN-1] Update URL anchor on mobile load

YasielCabrera:
## Ticket

## Description
When the finances page load on mobile resolution the URL anchors of the breakdown table were adding to the URL

## What solved
- [X] Finances view. Breakdown Table section. Removing filters reflected in the URL as parameters.-** **Expected Output:** The view should redirect to the first section and the URL should not display the filters.  **Current Output:** The filters are still visible in the URL even if the filters are not changed from the default state.


alejandrocabriales (approved):
LGTM

PR #1326, opened 2024-02-27
[BS-1] Fix/issues table number

alejandrocabriales:
## Ticket

## Description
Fix the order of of the metrics to match with the analiticys

## What solved
- [X] - Finances view. Breakdown Table section. Period: Annually. Year: 2024. - ** **Expected Output:** The total of each metric should match the sum of the value per each budget. **Current Output:**  The total of the Budget m metric is 20,138,275 but all the categories and subcategories have a value 0. All the totals displayed are wrong.


YasielCabrera (approved):
lgtm

PR #1327, opened 2024-02-27
Fix/percent in doughnut chart

alejandrocabriales:
## Ticket

## Description
Fix the % when the  Net Protocol Outflow is selected

## What solved
- [X] - Finances view, Legend section.Metric: Net Protocol Outflow- ** **Expected Output:** If the budgets have values, then the percent should be reflected. **Current Output:** MakerDao Legacy Budget displays a value of 36.5K which is reflected in the pie chart but the percent is 0


YasielCabrera (approved):
lgtm

PR #1329, opened 2024-02-27
Fix/fix height water chart

alejandrocabriales:
## Ticket

## Description
Remove the hover and correct size of the waterfall chart to show its legend

## What solved
- [X]  The legend info should be fully displayed. **Current Output:** The years are not fully displayed, because they are cut off. 
- [X]  The hover functionality shouldn't be displayed.


YasielCabrera (approved):
lgtm

PR #1331, opened 2024-02-27
[BSN-1] Fix mobile resize changes url

YasielCabrera:
## Ticket

## Description
When the finances page viewport is resized the we were adding the metrics, period and hash to the url

## What solved
- [X] Finances, Breakdown Chart section. URL parameters. Changing the screen size from mobile to desktop resolutions- ** **Expected Output:** The metrics should take the values depending on the resolution. **Current Output:** On mobile, a default metric is selected, but when switching to desktop, the metrics filter continues with a metric selected and the reset filter option is active.


alejandrocabriales (approved):
LGTM

PR #1345, opened 2024-03-04
fix:Fix the fontSize start and end bar of waterfall chart

alejandrocabriales:
## Ticket

## Description
Fix the font size in start and bar of the waterfall chart .
The actual font size of the waterfall es  to small so its need to be change

## What solved
- [X]  Finances,Reserves Chart section.- ** **Expected Output:** The numbers of each bar should be visible with a proper size.**Current Output:**  The numbers on the start and finish bars are barely visible.


YasielCabrera (approved):
lgtm

PR #1346, opened 2024-03-04
fix: Fix the loading state in breakdow chart

alejandrocabriales:
## Ticket

## Description
Fix the loading state in the breakdown chart in the Finances page .
The loading state got the wrong size in their bars, because the media query take time get correct value.
 The fix pass the items to Css to fix the problem

## What solved
- [X]  The loading state should be visible as the design shows for mobile resolution. **Current Output:**  It displays the loading state for the desktop and later changes to the proper resolution.


YasielCabrera (approved):
lgtm

PR #1348, opened 2024-03-05
fix:Fix the percent in mobile show diferent value

alejandrocabriales:
## Ticket

## Description
 Fix the  percent value in mobile that show incorrect percent for the same values.
 The percent in mobile is different that the one that is show in  desktop even although, values to calculate it are the same

## What solved
- [X] - The percent reflected in each budget/category should be the same in all resolutions.**Current Output:**  In the desktop resolution, the core units category displays correctly 99% but in mobile is displayed 98% for this category.


YasielCabrera (approved):
lgtm

PR #1349, opened 2024-03-05
fix: Fix the number of items show in the columns

alejandrocabriales:
## Ticket

## Description
Fix the correct items in each columns when there are to many items .
Some columns have more items that the other column, when should be the same  number

## What solved
- [X]- Finances-> MakerDAO Legacy Budget->core-units. Legend- ** **Expected Output:** All elements of the legend should be displayed properly covering the space. **Current Output:** In the second column a space without elements even when there are more elements to display.


YasielCabrera (approved):
lgtm

PR #1351, opened 2024-03-05
[BSN-1] Fix negative zero

YasielCabrera:
## Ticket

## Description
Fix negative zeros on the breakdown table

## What solved
- [X] **- Finances->MakerDAO Legacy Budget, Breakdown Table section. Year: 2023. Metric: Net Protocol Outflow - ** **Expected Output:** When the value is 0 the minus sign should not be displayed. **Current Output:** In the metric selected the PE-001 core unit displays -0.

### DIFF
--- a/src/core/utils/humanization.ts
+++ b/src/core/utils/humanization.ts
@@ -27,12 +27,13 @@ export const threeDigitsPrecisionHumanization = (num = 0, isHasAbsoluteValue = f
 export const usLocalizedNumber = (num: number, decimalPlace = 0): string => {
   num = num === 0 ? 0 : num; // avoid "-0"
 
-  return num?.toLocaleString('en-US', {
+  const value = num?.toLocaleString('en-US', {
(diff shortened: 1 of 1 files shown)


src/core/utils/humanization.ts
```diff
@@ -27,12 +27,13 @@ export const threeDigitsPrecisionHumanization = (num = 0, isHasAbsoluteValue = f
 export const usLocalizedNumber = (num: number, decimalPlace = 0): string => {
   num = num === 0 ? 0 : num; // avoid "-0"
```
alejandrocabriales, line 28: Can you remove the comment, because the line don't avoid the -0 value

alejandrocabriales (approved):

alejandrocabriales (changes requested):
Its possible to remove the comments?

alejandrocabriales (approved):
LGTM

PR #1352, opened 2024-03-05
fix:Fix the last level card overview don't show values

alejandrocabriales:
## Description
Fix the values in the last level of budgets.
Right now the last level of the budgets don't show correct values, because there aren't any subbudget but its have values that need to be show

## What solved
- [X] - Finances-> MakerDAO Legacy Budget->core-units->SES-001, First section. - ** **Expected Output:** The Actuals and Budget values should be reflected on the left side of the section. **Current Output:**  The first section shows both metrics with 0 values but the metrics line chart section displays values for both.


YasielCabrera (approved):
lgtm

PR #1354, opened 2024-03-06
[SB-1]  Fix the last level the doughnut and breakdown chart don't

alejandrocabriales:
…correct name

## Description
Fix the name in the last level for breakdown chart and drought chart Some of the  code path use to match came with some * in the code path so its not match

## What solved
- [X] The name of the budget/category/core unit should be the same in the sections. **Current Output:** In the pie chart legend is displayed "No-code" and the breakdown chart legend displays "No-code".


YasielCabrera (approved):
lgtm

PR #1355, opened 2024-03-06
[BS-1] Fix table don't show values in the last level in the header

alejandrocabriales:
## Ticket

## Description
Fix the values in the header  for the last level
Right now the  in the last level the table don't show correct values because don't have any subBudget

## What solved
- [X]  The table, at least the header should display the values of the metrics shown in the previous level (core units). **Current Output:**  In the core units level are displayed values for the TECH-00 but in its level, the table displays 0 in all metrics.


YasielCabrera (approved):
lgtm

PR #1363, opened 2024-03-08
[SB-1] Fix the filter waterfall chart

alejandrocabriales:
## Ticket

## Description
Fix the filter in the waterfall chart has the  wrong number to start The values to start the chart its only correct when all the budgets are select, The filter don't recalculate

## What solved
- [X] - Finances, Reserves Chart section ** **Expected Output:** It should display the proper data corresponding to the budget/core unit selected. The general data is always displayed.


YasielCabrera (approved):
lgtm

PR #1366, opened 2024-03-11
[SB-1]Fix sum values of the items in the  pie chart is not 100%

alejandrocabriales:
## Ticket

## Description
Fix the sum of the all values in the pie chart and the mobile view should be 100%.
Right now because the round some times the sum of all items its less or more than 100%

## What solved
- [X] All.  **Resolution:** All **Steps to Reproduce:** **- Finances->MakerDAO Legacy Budget-> core units.- ** **Expected Output:** The sum of the percent should have as result 100%.


YasielCabrera (approved):
lgtm

PR #1372, opened 2024-03-12
[SB-1] Fix  values from the table are missing

alejandrocabriales:
## Ticket

## Description
Fix the values in the table that are  no values
When there are  only one subBuget the same budget are add twice

## What solved
- [X] :- Finances->MakerDAO Legacy Budget-> SPFs. Breakdown Table.- ** **Expected Output:** The data relating to the metrics for each budget/ category in the first should be reflected in the rest of the sections.  **Current Output:** The first section displays values for the Budget metric but they are not reflected in the breakdown table.


YasielCabrera (approved):
lgtm

PR #1377, opened 2024-03-12
[SB-1] Avoid reset budget filter in waterfall chart

alejandrocabriales:
## Ticket

## Description
Avoid reset filter of budgets, when the granularity change

## What solved
- [X] - Finances view. Reserves Chart section, All Categories filter. - ** **Expected Output:** The categories/budgets selected should be maintained when the period filter is changed.   **Current Output:**  Changing the period, the categories are reset to the default state


YasielCabrera (approved):
lgtm

PR #1378, opened 2024-03-13
[SB-1] Reset category filter to initial state

alejandrocabriales:
## Ticket

## Description
Add the default value for categories in the waterfall chart

## What solved
- [X] - Finances, Reserves Chart section. Changing the period filter and selecting the Reset option.- ** **Expected Output:** Both filters should back to their default state. **Current Output:** All Categories is reset but without all elements selected.


YasielCabrera (approved):
lgtm

PR #1385, opened 2024-03-14
[SB-1] Fix when only one metric is select show the name of the metric

alejandrocabriales:
## Ticket

## Description
When one metric is select only show the same placeholder. When only one metric is select then should be show the  metric instead of placeholder with number of items

## What solved
- [X] BSN-1: (applicable to all budget levels): breakdown table: when it only has 1 metric selected it should show the name of that metric, and if it has multiple it should show the number between parentheses: Metrics (2).


YasielCabrera (approved):
lgtm

PR #1390, opened 2024-03-16
[QA] Move code below title

YasielCabrera:
## Ticket

## Description
Move the code below the title if the code is lower case, has dashes and it is not uppercase, or if it is in camelCase format

## What solved
- [X] Budget path as a code and/or long codes,  the code should be displayed below the name if they math the condition (lowercae | - (dash present) | camelCase).


alejandrocabriales (approved):
LGTM

PR #1396, opened 2024-03-19
[BSN-1] Fix when there its not analytics but there are budgets

alejandrocabriales:
## Ticket
[N/A]

## Description
[N/A]

## What solved
- [X] Fix when there its not analytics and exist budgets


YasielCabrera (approved):
lgtm

PR #1399, opened 2024-03-19
[QA-338] Filter element that have no cero values

alejandrocabriales:
## Ticket

## Description
Remove the items from the tooltip if the values are cero

## What solved
[X] Breakdown chart popup when hovering on a specific granularity bar should not show metrics that are 0. Applicable to all levels.


DianaTahiri (approved):
LGTM

PR #1406, opened 2024-03-20
[N/T] Add condition to show with cero values

alejandrocabriales:
## Ticket
N/A

## Description
Add the possibility to show cero values in metric tooltip

## What solved
- [X] Fix the tooltip don't show zero values


YasielCabrera (approved):
lgtm

PR #1411, opened 2024-03-22
[QA-338] Allow show name of metric when only one its select waterfall

alejandrocabriales:
## Ticket

## Description
Allow to show the name of budget when only one its selected and only show number metrics when its mobile

## What solved
- [X] Reserves Chart: when the All Categories are not selected, the "ALL"  should be removed. 
- [X] Reserves Chart: When one category is selected, the filter should display the category name (no icon), more than one should display Categories (#). In mobile, use always Categories (#) .


YasielCabrera (approved):
lgtm

PR #1416, opened 2024-03-22
[BSN] Change default amount of metrics on bigger desktops

YasielCabrera:
## Ticket
https://trello.com/c/FnTQ74C9/348-bsn-1-general-issues

## Description
From 2400px up we should 4 metrics and form 3000 on we show the 5 metrics by default

## What solved
- [X] Finances view, Breakdown Table section.- ** **Expected Output:** The amount of the metrics selected by default should depend on the screen resolution. **Current Output:** By default 3 metrics are selected but the table displays space that can be covered with another metric.


alejandrocabriales (approved):
LGTM

PR #1421, opened 2024-03-25
[BSN] Remove no-code work and put the path fragment to fix it

YasielCabrera:
## Ticket

## Description
using a name extracted by the path if there is not name or code

## What solved
- [X] Finances view, Breakdown Table section. Clicking the atlas/legacy/scopes sub-budget.- ** **Expected Output:** The subcategories name should match in all sections. **Current Output:** The legend shows subcategories with no-code name. The breakdown Chart displays them as Not name and the Reserves chart shows No-key-.


alejandrocabriales (approved):
LGTM

PR #1424, opened 2024-03-26
[BSN-348] Add the umbral condition for start point

alejandrocabriales:
## Ticket

## Description
Avoid big negative number with decimal, replace with cero, for the start point of the waterfall

## What solved
- [X]Finances->MakerDAO Legacy Budget->Keepers: http://46.101.104.232:82/finances/legacy/keepers?year=2023. Reserves Chart section.- ** **Expected Output:** If the Start column has the value of 0, then the number visible should be 0. **Current Output:**  The Start column is displayed with value = - 0.00267.


YasielCabrera (approved):
lgtm

PR #1426, opened 2024-03-26
[BSN] Fix waterfall filter scroll

YasielCabrera:
## Ticket

## Description
Remove scroll on filter filed popups

## What solved
- [X] Breakdown chart and Breakdown Table sections. Metric filter. **Expected Output:** All metrics should be shown and the scroll bar should not be visible. **Current Output:** All metrics are displayed but the scroll bar appears.


alejandrocabriales (approved):
LGTM

PR #1434, opened 2024-03-28
[BUG-366] Take the zone from the time zone client

alejandrocabriales:
## Ticket
https://trello.com/c/wxhA1BDB/366-bug-no-access-to-march-2024-in-the-ecosystem-actors-expense-reports

## Description
Should allow navigating through to March 2024.

## What solved
- [X] Should allow navigating through to March 2024.


YasielCabrera (changes requested):
the timezone should not be the same of the client

PR #1437, opened 2024-03-29
Hide zero value son the breakdown chart

YasielCabrera:
## Ticket

## Description
Hide close to 0 values on the breakdown chart

## What solved
- [X] Finances. Breakdown Chart section. Metric: Net Expenses On-chain- ** **Expected Output:** The tooltip should show only the metrics with a value >0.  **Current Output:**  The March 2023 column shows the bar with straight borders and the tooltip reflects the Scope Framework metric with a value = 0.

### DIFF
--- a/src/stories/containers/Finances/components/BreakdownChartSection/utils.ts
+++ b/src/stories/containers/Finances/components/BreakdownChartSection/utils.ts
@@ -81,7 +81,8 @@ export const setBorderRadiusForSeries = (
 
     // Identify first and last indices for positive and negative values
     series.forEach((s, seriesIndex) => {
-      const value = s.data[dataIndex].value ?? 0;
+      let value = s.data[dataIndex].value ?? 0;
+      value = Math.abs(value) < 0.004 ? 0 : value; // if the values es too small, consider it as 0
       if (value > 0) {
         if (firstPositiveIndex === -1) firstPositiveIndex = seriesIndex;
         lastPositiveIndex = seriesIndex;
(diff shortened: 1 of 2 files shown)


src/stories/containers/Finances/components/BreakdownChartSection/utils.ts
```diff
@@ -81,7 +81,8 @@ export const setBorderRadiusForSeries = (
 
     // Identify first and last indices for positive and negative values
     series.forEach((s, seriesIndex) => {
-      const value = s.data[dataIndex].value ?? 0;
+      let value = s.data[dataIndex].value ?? 0;
+      value = Math.abs(value) < 0.004 ? 0 : value; // if the values es too small, consider it as 0
```
alejandrocabriales, line 85: I am in doubt if 0.003 will work, can we use something like this ?
value = (value > 0 && value < 1) ? 0: value;
Let me know you opinion

alejandrocabriales (commented):
LGTM

src/stories/containers/Finances/components/BreakdownChartSection/utils.ts
YasielCabrera: Note that the issue to be solved here is that when the values is too small it is showing `0.00`. So if it is at least `0.01` it can be shown in the UI, if the absolute values is > 0.004 it means that it can be rounded up to 0.01 or -0.01 so it can be shown, if it is less than that value we can assume it as 0. That's why it can't be compared to be between 0 and 1

alejandrocabriales: LGTM

PR #1439, opened 2024-03-29
Fix expense reports metric label

YasielCabrera:
## Ticket

## Description
Fixed the metric label on small resolutions
## What solved
- [X] Expense Reports section. Metric: Net Protocol Outflow. **Expected Output:**  The name displayed in the table should be similar to the filter's label as much as possible. **Current Output:** The filter's label displays Net Protocol Outflow but in the table, it refers to Protocol Net Outflow.  The text is too close to the Status label.


alejandrocabriales (approved):
LGTM

PR #1442, opened 2024-03-31
Improve budget pager dates

YasielCabrera:
## Ticket
https://trello.com/c/wxhA1BDB/366-bug-no-access-to-march-2024-in-the-ecosystem-actors-expense-reports

## Description
Change the dates to be UTC localized to avoid differences between server/client dates

## What solved
- [X] Should allow navigating through to March 2024.


alejandrocabriales (approved):
lgtm

PR #1448, opened 2024-04-02
Fix metrics flickering on period change on breakdown table

YasielCabrera:
## Ticket

## Description
Fix metric flickering when the amount of metric to be shown is bigger than the previous amount of metric selected previously with the previous period

## What solved
- [X] **Steps to Reproduce:** Breakdown Table section. Changing the period from Annually to Semi-annual.  **Expected Output:**  If the data takes time to be loaded, the loading state should be displayed. **Current Output:** The header displays the values of the three metrics and shows the proper metric.


alejandrocabriales (commented):
LGTM

PR #1457, opened 2024-04-03
Add missing path data to breakdown table

YasielCabrera:
## Ticket

## Description
Add missing values to the table when there are path in the analytics that doesn't match with any budget

## What solved
- [X] Finances->Scope Framework Budget->Support Scope, Pie chart section. Metric: Net Expenses On-chain.  **Expected Output:**  The data for each metric should be reflected in all sections. **Current Output:** Pie chart legend displays 13 items, Incubation has 100% of the value but the navigation cards section shows only 12 cards. The breakdown table doesn't display any value for the Net Expenses On-chain metric selected.


alejandrocabriales (approved):
LGTM

PR #1458, opened 2024-04-04
[QA] Limit breakdown table rows

YasielCabrera:
## Ticket

## Description
Allow max 13 rows per sub table

## What solved
- [X] If there are more than 12 rows, group all additional rows (from the 13th row onwards) under a single line item named "Others". Others =/= Other, the latter is an uncategorized budget, while the former is a grouping of budget line items.  This rule applies to the specific Budget Category within a Breakdown Table and not to the Breakdown Table as a whole. Since currently the "Others" rule is applied to the Breakdown Table as a whole, the rest of the Budget Categories have no children...


alejandrocabriales (approved):
LGM

PR #1459, opened 2024-04-05
[QA] Empty doughnut charts in the last levels

YasielCabrera:
## Ticket

## Description
Add empty values on levels where there are not data in the doughnut chart

## What solved
- [X] When there is a single budget entity like NGT NewGov Token or AA AllocatorDAO Axon, make it visible in the donut section. NGT example is showing the correct way. Applicable to all instances of singular budget entities.
- [X] Deepest level: let's make it to the right of the donut, legend icon, gray if there is no value:


alejandrocabriales (approved):
LGTM

PR #1466, opened 2024-04-08
Fix last level doughnut chart color

YasielCabrera:
## Ticket
https://trello.com/c/s9JOxwZX/368-change-requests-sprint-review-v3

## Description
Make the last level with only one item and zero values gray

## What solved
- [X] **The change should be done in the legend icon instead of the pie chart.** Deepest level: let's make it to the right of the donut, legend icon, gray if there is no value


alejandrocabriales (approved):
lgtm

PR #1467, opened 2024-04-08
Fix waterfall chart finish date

YasielCabrera:
## Ticket

## Description
The finish date should be next year instead the same year

## What solved
- [X] Reserves Chart: Finish and start reserves should match.


alejandrocabriales (approved):
LGTM

PR #1471, opened 2024-04-09
Add the code to code and fixscrollbar ecosystem actor filter

alejandrocabriales:
## Ticket

## Description
Fix the issues relate with the ecosystem actors filters

## What solved
- [X] Should the empty state be handled when there is no result.
- [X] The entire name including the code should be displayed.
- [X]  When all the elements of the component are displayed, a scroll bar is not needed.

## Screenshots (if apply)
testing


YasielCabrera (approved):
lgtm

PR #1474, opened 2024-04-09
[GI-371] Move the conditional to css for better user experience

alejandrocabriales:
## Ticket
https://trello.com/c/ol1mo4dH/371-finances-view-general-issues-v3

## Description
This add the conditional render to css, for better user experience

## What solved
- [X] Finances. The first section (left side). **Expected Output:** Refreshing the page The Net Exp On-chain and Budget Cap labels should be loaded properly. **Current Output:** The Net Expenses On-chain label is loaded and then it is changed to the Net Exp On-chain.


YasielCabrera (approved):
lgtm

PR #1476, opened 2024-04-10
Fix doughnut percentages

YasielCabrera:
## Ticket
https://trello.com/c/ol1mo4dH/371-finances-view-general-issues-v3

## Description
If the values for some items are too small relative to the others, the percentage was close to 0 but it was showing 0%

## What solved
- [X] Finances view, Pie chart section. Metric: Net Expenses On-chain.  **Expected Output:** The percent should be based on the values displayed by each budget. **Current Output:** Atlas Immutable displays 16.0k but with 0%


alejandrocabriales (approved):
LGTM

PR #1477, opened 2024-04-10
[QA] Fix mobile cards budget utilization

YasielCabrera:
## Ticket

## Description
Change the percentage of the budget utilization in the cards on mobile and change the budget by net on chain value

## What solved
- [X] mobile only / since there is no donut view, lets indicate budget utilization for each subbudgets; so we are not (on mobile only) going to show how much a subbudget is a part of the larger budget, but we are going to show a budget utilization of each specific sub budget. Budget boxes - metric shown should be utilization number - applicable to all levels on mobile.


alejandrocabriales (approved):
LGTM

PR #1490, opened 2024-04-11
[QA-376] Add condition when values are cero don't show values

alejandrocabriales:
## Ticket
https://trello.com/c/9cI0ENhT/376-qa-issues-bug-findings-fusion-v4

## Description
This remove the row when the values are cero

## What solved
- [X] Breakdown table section. Uncategorized row. **Expected Output:** The line item should not be displayed if the value is 0.  **Current Output:** The line is displayed even if the value is 0.


YasielCabrera (approved):
lgtm

PR #1550, opened 2024-05-08
[QA] Fix uncategorized rules

YasielCabrera:
## Ticket
https://trello.com/c/iDmNeoEO/434-qa-item

## Description
ALL SCREENS // Uncategorized should have no children. 1. If the uncategorized line item is the only one for the table or subsections of the table, then hide it and include it in the totals row for that specific table and/or subsection of the table. 1.1. So if Uncategorized line item is the only one for e.g., Support Scope, then include the values of the uncategorized in the Support Scope totals row. 2. If uncategorized is not the only one then show it with other line items. 2.1. If Support Scope has other line items on top of Uncategorized: it has e.g., Uncategorized, Incubation budget etc..., then show these all as line items under Support Scope. 3. Uncategorized as a line item subsumes every subline item, i.e., it does not have any children.


alejandrocabriales (approved):
LGTM

PR #1551, opened 2024-05-09
[No- ticket] Add check when there its no rows[0]

alejandrocabriales:
## Ticket
No ticket

## Description
Fix the problem  in the uncategorized in the legacy

## What solved
- [X] Fix the problem  in the uncategorized in the legacy


YasielCabrera (approved):
lgtm

PR #1553, opened 2024-05-10
[Fusion] Apply reskin design to the Key Changes section (MET)

alejandrocabriales:
## Ticket
https://trello.com/c/04eKxJok/426-apply-reskin-design-to-the-key-changes-section-met

## Description
This add a res-kin to the  Key Changes sections

## What solved
- [X] Should add the image background with the new styles.
- [X] Should update the background image.
- [X] Should update the description with the new styles. 
- [X] Should create the button for internal views (state: default, hover, click).


src/components/InternalLinkButton/InternalLinkButton.tsx
```diff
@@ -0,0 +1,88 @@
...
+const InternalLinkButton: React.FunctionComponent<Props> = ({
+  label,
+  icon,
+  href,
+  className,
+  iconFill = '#5B667E',
+  iconWith = 24,
+  iconHeight = 23,
+  background,
+  iconClassName,
```
YasielCabrera, line 27: I recomment to rename the prop `icon` to `showIcon` or something similar to indicate with its name that the prop is a boolean. Also the props `iconFill` should not be there as we can change the `fill` by overriding the styles thanks to the `className` using a selector like `& svg`. the other props: `iconWidth`, `iconHeight` should also be handled using the selector like the fill, the `background` should be applied the same logic otherwise we'll need to do something like `background={isLight ? 'x' : 'y'}` while with CSS selector we can use pure CSS to handle the theme.
Same with `iconClassName`, with this we need to generate the styles outside, get the className and pass to it and if I'm not wrong that is deprecated in MUI

src/components/InternalLinkButton/InternalLinkButton.tsx
```diff
@@ -0,0 +1,88 @@
...
+}) => (
+  <Link href={href ?? ''} passHref legacyBehavior>
```
YasielCabrera, line 29: We should not use the `legacyBehavior` anymore, not in next >13, the `Link` component is rendered as a `a` tag so we don't need to add a nested `a` tab below. You can remove the `Container` and pass the logic to the styling of the `Link` itself

src/components/InternalLinkButton/InternalLinkButton.tsx
```diff
@@ -0,0 +1,88 @@
...
+  background:
+    background || (theme.palette.isLight ? theme.palette.colors.gray[100] : theme.palette.colors.charcoal[800]),
+  gap: 8,
+
+  ':hover': {
+    gap: 16,
+    padding: '4px 8px 4px 24px',
+    border: theme.palette.isLight
+      ? `1px solid ${theme.palette.colors.gray[300]}`
+      : `1px solid ${theme.palette.colors.charcoal[100]}`,
```
YasielCabrera, line 59: if you add border on hover you will see a little jump in the component when it's hovered, this is because the hover is added to the space, to prevent this you can add a transparent border with the same width when the component is not hovered or just reduce 1px to the padding when hovered as the border has a 1px width in this case.

src/stories/components/svg/ArrowInternalLink.tsx
```diff
@@ -0,0 +1,27 @@
+import * as React from 'react';
+
+interface Props {
+  width?: number;
+  height?: number;
+  className?: string;
+  fill?: string;
+}
+
+const ArrowInternalLink: React.FC<Props> = ({ className, width = 24, height = 24, fill = '#5B667E', ...props }) => (
```
YasielCabrera, line 10: this is a new component so it should be in the new component folder

src/stories/containers/Endgame/EndgameContainer.tsx
```diff
@@ -7,7 +7,7 @@ import { toAbsoluteURL } from '@ses/core/utils/urls';
 import React from 'react';
 import BudgetStructureSection from './components/BudgetStructureSection/BudgetStructureSection';
 import BudgetTransitionStatusSection from './components/BudgetTransitionStatusSection/BudgetTransitionStatusSection';
-import EndgameIntroductionBanner from './components/EndgameIntroductionBanner/EndgameIntroductionBanner';
+import EndgameIntroductionBannerReskin from './components/EndgameIntroductionBannerReskin/EndgameIntroductionBannerReskin';
```
YasielCabrera, line 10: Instead of using the word `Reskin`, use `legacy` in the old component, as this component is the final implementation

src/stories/containers/Endgame/EndgameContainer.tsx
```diff
@@ -63,8 +63,8 @@ const EndgameContainer: React.FC<EndgameContainerProps> = ({ budgetTransitionAna
       </Container>
 
       <div ref={keyChangesRef}>
-        <BannerContainer id="section-key-changes">
-          <EndgameIntroductionBanner isKeyChanges />
+        <BannerContainer>
```
YasielCabrera, line 66: Please, note that if you remove the `id` the navigation is not going to work properly

src/stories/containers/Endgame/components/EndgameIntroductionBannerReskin/EndgameIntroductionBannerReskin.stories.tsx
```diff
@@ -0,0 +1,100 @@
+import { createThemeModeVariants } from '@ses/core/utils/storybook/factories';
+import EndgameIntroductionBannerReskin from './EndgameIntroductionBannerReskin';
+import type { Meta } from '@storybook/react';
+import type { FigmaParams } from 'sb-figma-comparator';
+
+const meta: Meta<typeof EndgameIntroductionBannerReskin> = {
+  title: 'Components/Endgame/Endgame Introduction Banner Reskin',
```
YasielCabrera, line 7: Similar to the previous comment about the reskin work, it should be removed here or replaced by legacy if this has the legacy styles

YasielCabrera (changes requested):

src/components/InternalLinkButton/InternalLinkButton.tsx
alejandrocabriales: you are right, done

src/components/InternalLinkButton/InternalLinkButton.tsx
alejandrocabriales: Good to know , done

src/components/InternalLinkButton/InternalLinkButton.tsx
alejandrocabriales: you are right, done

src/stories/components/svg/ArrowInternalLink.tsx
alejandrocabriales: you are right, done

src/stories/containers/Endgame/EndgameContainer.tsx
alejandrocabriales: you are right, done

src/stories/containers/Endgame/EndgameContainer.tsx
alejandrocabriales: you are right, done

src/stories/containers/Endgame/components/EndgameIntroductionBannerReskin/EndgameIntroductionBannerReskin.stories.tsx
alejandrocabriales: you are right, done

YasielCabrera (approved):
lgtm

PR #1556, opened 2024-05-14
[Reskin-432]Add new button On-Chain and Budget Cap and new font for the chart

alejandrocabriales:
## Ticket
https://trello.com/c/XToNKJ7T/432-apply-reskin-design-to-the-budget-transition-status-section-met

## What solved
- [X] FR**(3.5)** Should create and update the Net Expenses On-Chain and Budget Cap buttons (status: default, hover, click).


styles/globals.scss
```diff
@@ -128,6 +128,27 @@
...
+  font-style: normal;
+  font-display: swap;
+}
+@font-face {
+  font-family: 'OpenSansCondensed';
+  src: url('../assets/fonts/OpenSansCondensed/OpenSansCondensed-LightItalic.ttf');
+  font-weight: 300;
+  font-style: normal;
+  font-display: swap;
+}
```
YasielCabrera, line 151: both fonts declarations have the same attributes but they're actually different, maybe the last one should be ` font-style: italic`

YasielCabrera (changes requested):

styles/globals.scss
alejandrocabriales: you are right, done

YasielCabrera (approved):
lgtm

PR #1568, opened 2024-05-16
[Reskin-432] Add style for mobile view in tooltip add deprecate for legacy component

alejandrocabriales:
## Ticket
https://trello.com/c/XToNKJ7T/432-apply-reskin-design-to-the-budget-transition-status-section-met

## What solved
- [X] Should update the tooltip with the copy when hover over the bars.
- [X] Should apply the hover over the bars.

### DIFF
--- a/src/components/SESTooltip/TooltipModalVariant.tsx
+++ b/src/components/SESTooltip/TooltipModalVariant.tsx
@@ -1,5 +1,4 @@
-import { styled } from '@mui/material';
-import { useThemeContext } from '@ses/core/context/ThemeContext';
+import { styled, useTheme } from '@mui/material';
 import BasicModal from '@/stories/components/BasicModal/BasicModal';
 import { Close } from '@/stories/components/svg/close';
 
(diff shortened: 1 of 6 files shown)


src/components/SESTooltip/TooltipModalVariant.tsx
```diff
@@ -9,7 +8,8 @@ interface TooltipModalVariantProps extends React.PropsWithChildren {
 }
 
 const TooltipModalVariant: React.FC<TooltipModalVariantProps> = ({ children, openModal, handleCloseModal }) => {
-  const { isLight } = useThemeContext();
+  const theme = useTheme();
```
YasielCabrera, line 11: You can use a callback in `sx` props to use the `theme` so you don't need this. Take a look in the [docs](https://github.com/makerdao-ses/ecosystem-dashboard/pull/1569)

YasielCabrera (changes requested):

src/components/SESTooltip/TooltipModalVariant.tsx
alejandrocabriales: you are right , done

YasielCabrera (approved):
lgtm

PR #1572, opened 2024-05-17
[QA-429]: fix issue with the navigation tab on the Endgame page

yoelvyspc93:
## Ticket
https://trello.com/c/ki5J9mbH/429-endgame-budget-structure-item-is-not-highlighted-on-the-endgame-page

## What solved
- [X] Should the navigation item be highlighted when it navigates to each section of the page.


YasielCabrera (changes requested):
Note that each time a section is changed due to a click on the nav or a scroll into the section, the URL should be updated with the ID of the section as it is a requirement to allow the users to share the URL to a specific section. Also, when the page is opened/loaded it should scroll to the section

alejandrocabriales (approved):
LGTM

PR #1573, opened 2024-05-18
[Resking-441]Add the style for the text and status Chip component

alejandrocabriales:
## Ticket
https://trello.com/c/MJ9L8XdX/441-apply-reskin-design-to-the-ea-index

## What solved
- [X] Should update the view title with the new styles
- [X] Should update the view description with the new styles.
- [X] Should update the Filter components for Desktop resolutions (status: default, hover, pressed)


src/components/StatusChip/StatusChip.stories.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { StatusChip } from './StatusChip';
+import type { Meta } from '@storybook/react';
+
+const meta: Meta<typeof StatusChip> = {
+  title: 'Fusion/Components/StatusChip',
+  component: StatusChip,
+
+  parameters: {
+    chromatic: {
+      viewports: [375, 768, 1024, 1280, 1440, 1920],
```
YasielCabrera, line 11: This component looks to be the same in all the resolutions so we don't need to specify the `viewports` for this case

src/stories/components/StatusChipLegacy/StatusChipLegacy.tsx
```diff
@@ -54,14 +54,16 @@ const colors: { [id: string]: any } = {
     backgroundDark: '#FF8E3633',
   },
 };
-
-export const StatusChip = (props: StatusChipProps) => {
+/**
+ * @deprecated use StatusChip  instead
```
YasielCabrera, line 58: Are those two statuses the same thing? the legacy status can be 4 variants only and this includes "Obsolete" which is not a budget statement status. Maybe we should  handle the new one as other thing (including a different name if they're different) (Note that the new one looks like a CU/EA status

src/stories/containers/ActorsAbout/components/ActorSummary/ActorTitleAbout.tsx
```diff
@@ -2,7 +2,7 @@ import styled from '@emotion/styled';
 import { Typography, useMediaQuery } from '@mui/material';
 import { CircleAvatar } from '@ses/components/CircleAvatar/CircleAvatar';
 import SocialMediaComponent from '@ses/components/SocialMediaComponent/SocialMediaComponent';
-import { StatusChip } from '@ses/components/StatusChip/StatusChip';
+// import { StatusChip } from '@ses/components/StatusChip/StatusChip';
```
YasielCabrera, line 5: We can remove this commented code

YasielCabrera (changes requested):

src/components/StatusChip/StatusChip.stories.tsx
alejandrocabriales: you are right, done

src/stories/containers/ActorsAbout/components/ActorSummary/ActorTitleAbout.tsx
alejandrocabriales: you are right, done

src/stories/components/StatusChipLegacy/StatusChipLegacy.tsx
alejandrocabriales: you are right, done

YasielCabrera (approved):
lgtm

PR #1579, opened 2024-05-21
[MET] Update key result card header

YasielCabrera:
## Ticket
https://trello.com/c/04eKxJok/426-apply-reskin-design-to-the-key-changes-section-met

## Description
Updated the Ket results card header and the hover of the accordions

## What solved
- [X] Should apply the hover state for the accordion sections.


alejandrocabriales (approved):
LGTM

PR #1592, opened 2024-05-23
[Reskin-446]: new select component

yoelvyspc93:
## Ticket
https://trello.com/c/UGndl7pm/446-create-a-reusable-year-component-filtersbundle-for-the-endgame-view

## What solved
- [X] Should create the year component with the default, hover, and pressed statuses.
- [X] Should create the year list (single select) with the hover, selected statuses. Desktop resolutions.
- [X] The years' list should have color

### DIFF
--- a/src/components/CustomSelect/CustomSelect.tsx
+++ b/src/components/CustomSelect/CustomSelect.tsx
@@ -1,7 +1,7 @@
 import { ExpandMore, Check } from '@mui/icons-material';
 import { Select, MenuItem, FormControl, styled, Box, Typography, useTheme } from '@mui/material';
 import React from 'react';
-import type { CustomSelectProps } from './type';
+import type { CustomSelectProps, OptionItem } from './type';
 import type { SelectChangeEvent, Theme } from '@mui/material';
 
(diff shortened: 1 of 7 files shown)


src/components/CustomSelect/CustomSelect.tsx
```diff
@@ -0,0 +1,205 @@
+import { ExpandMore, Check } from '@mui/icons-material';
+import { Select, MenuItem, FormControl, styled, Box, Typography, useTheme } from '@mui/material';
+import React from 'react';
+import type { CustomSelectProps } from './type';
+import type { SelectChangeEvent, Theme } from '@mui/material';
+
+const CustomSelect: React.FC<CustomSelectProps> = ({
+  label,
```
alejandrocabriales, line 9: What do you think to pass className?:string, for add new styles if its need in others places??

yoelvyspc93: in the new update it has been implemented

alejandrocabriales: nice, done

PR #1626, opened 2024-05-31
[Fix-450] implement the funtionality for the filters

yoelvyspc93:
## Ticket

## What solved
- [X] Should implement the funtionality for the filters


alejandrocabriales (approved):
lgtm

PR #1628, opened 2024-06-01
[Update-453] Add match page actors with figma

alejandrocabriales:
## Ticket
https://trello.com/c/WXdC6g15/453-update-the-storybook-structure

## What solved
- [X] Should integrate the Endgame page with the Figma image.

### DIFF
--- a/src/components/FiltersBundle/FilterDesktop.tsx
+++ b/src/components/FiltersBundle/FilterDesktop.tsx
@@ -64,7 +64,7 @@ const FilterDesktop: React.FC<FilterDesktopProps> = ({ filters, searchFilter, re
         <CustomSearch
           placeholder="Search"
           onChange={searchFilter.onChange}
-          style={searchFilter.style}
+          style={searchFilter.widthStyles}
           value={searchFilter.value}
         />
       </SearchWrapper>
--- a/src/components/FiltersBundle/FilterDesktop.tsx
+++ b/src/components/FiltersBundle/FilterDesktop.tsx
@@ -61,7 +61,7 @@ const FilterDesktop: React.FC<FilterDesktopProps> = ({ filters, searchFilter, re
     {!!searchFilter && (
       <SearchWrapper>
         <CustomDivider orientation="vertical" flexItem />
-        <CustomSearch placeholder="Search" onChange={searchFilter.onChange} />
+        <CustomSearch placeholder="Search" onChange={searchFilter.onChange} style={searchFilter.style} />
       </SearchWrapper>
     )}
   </FilterElement>
@@ -76,6 +76,7 @@ const ResetButton = styled(Button)(({ theme }) => ({
   lineHeight: '24px',
   borderRadius: 6,
   background: 'transparent',
+  // padding: '4px 14px',
   padding: '4px 16px',
   textTransform: 'none',
   '&:hover': {
@@ -101,12 +102,9 @@ const SearchWrapper = styled('div')({
   gap: 16,
 });
 
-const CustomSearch = styled(Search)({
+const CustomSearch = styled(Search)(() => ({
   display: 'flex',
   '& > div': {
     width: '100%',
   },
-  '& > input': {
-    backgroundColor: 'red',
-  },
-});
+}));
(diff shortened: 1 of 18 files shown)


src/views/Actors/components/ActorTable/ActorTable.tsx
```diff
@@ -36,9 +36,9 @@ const TableWrapper = styled.div({
 const ContainerList = styled.div({
```
YasielCabrera, line 36: We should migrate to the `@material/styles` instead of continue using @emotion/styles

src/views/Actors/components/ActorTable/ActorTable.tsx
```diff
@@ -36,9 +36,9 @@ const TableWrapper = styled.div({
...
   display: 'flex',
   flexDirection: 'column',
-  gap: 24,
+  gap: 8,
   [lightTheme.breakpoints.up('tablet_768')]: {
```
YasielCabrera, line 40: Once changed to `@material` styles, we should use `theme` instead of `lightTheme` here

src/views/Actors/useActors.tsx
```diff
@@ -380,10 +351,39 @@ export const useActors = (actors: Team[], stories = false) => {
...
+      label: 'Actor Role',
+      selected: filteredCategories,
+      multiple: true,
+
+      options: categoryOptions,
+      onChange: (value: string | number | (string | number)[]) => {
+        if (typeof value === 'string' || (Array.isArray(value) && value.every((item) => typeof item === 'string'))) {
+          handleChangeUrl('filteredCategories')(value as string | string[]);
+        } else {
+          console.error('Invalid value type: must be string or string array');
```
YasielCabrera, line 370: url params are string or arrays only, and you're already checking that, if the values are any other type, it is a programming error so we don't need to handle this console. It doesn't make too much sense from the user perspective anyway as it is a console print

src/components/FiltersBundle/FilterDesktop.tsx
```diff
@@ -76,6 +82,7 @@ const ResetButton = styled(Button)(({ theme }) => ({
   lineHeight: '24px',
   borderRadius: 6,
   background: 'transparent',
+  // padding: '4px 14px',
```
YasielCabrera, line 85: did you forgot the remove this commented code?

src/components/FiltersBundle/types.ts
```diff
@@ -18,6 +18,13 @@ export interface ResetFilter {
 
 export interface SearchFilter {
   onChange: (value: string) => void;
+  value?: string;
+  inputRef?: React.RefObject<HTMLInputElement>;
```
YasielCabrera, line 22: Why do we need a ref here? Everything related to input changes should be handled using the `value` and `onChange` properties

src/components/FiltersBundle/types.ts
```diff
@@ -18,6 +18,13 @@ export interface ResetFilter {
...
+  style?: {
```
YasielCabrera, line 23: we should name this property with other names to avoid confusion with the HTML attribute `style` as it is not behaving as the `style` property and it has specific values. Maybe `widthStyles` could work better?

src/views/Actors/components/ActorItem/ActorItem.tsx
```diff
@@ -792,3 +793,9 @@ const PopoverListLinksStyled = styled(PopoverListLinks)(({ theme }) => ({
     },
   },
 }));
+
+const ScopeChipStyled = styled(ScopeChip)(({ theme }) => ({
+  [theme.breakpoints.up('desktop_1024')]: {
+    height: 24,
+  },
+}));
```
YasielCabrera, line 801: Just a doubt, no change request here: is ScopeChip used only here or in other places too? Because if it is the same component it should have the styles inside the component itself and we can just use it here. It looks like it is a design issue or we're adding extra padding/font sizes or something and that's why, maybe, we need to adjust the height

YasielCabrera (changes requested):

src/views/Actors/components/ActorTable/ActorTable.tsx
alejandrocabriales: done

src/views/Actors/useActors.tsx
alejandrocabriales: done

src/views/Actors/components/ActorItem/ActorItem.tsx
alejandrocabriales: Yes , its the same component, inside of this view the component have different style

src/views/Actors/components/ActorTable/ActorTable.tsx
alejandrocabriales: done

src/components/FiltersBundle/types.ts
alejandrocabriales: done

src/components/FiltersBundle/types.ts
alejandrocabriales: done

YasielCabrera (approved):
lgtm

PR #1630, opened 2024-06-03
[T-450] Fix reset button statuses

YasielCabrera:
## Ticket

## Description
Show the reset filter button with the enable/disable statuses. Also renamed the actors container to ActorsView to follow the naming convention

## What solved
- [X]  **Expected Output:** The Reset Filter option should be active when at least one element is selected from the filter or text has been inserted on the Search component. **Current Output:** The Reset Filters option is never enabled.


alejandrocabriales (approved):
lgtm

PR #1637, opened 2024-06-04
[T-450] Fix EA filter labels

YasielCabrera:
## Ticket

## Description
Allow to show the full width of the select on the filter bundle on certain occasions and allow to show correctly the labels of the selects 

## What solved
- [X]  Scopes Filter. **Expected Output:** When more than one scope is selected, the filter label should be "Scopes (#)". **Current Output:** The label is cut off therefore the number is not visible. 
- [X]  Scopes and Actor Role filters. **Expected Output:** When an element is selected, the element name should appear in full in the filter. **Current Output:** The name is cut off


alejandrocabriales (approved):
lgtm

PR #1646, opened 2024-06-05
[Reskin-440] Center text in team members and match expenditure component

alejandrocabriales:
## Ticket

## What solved
- [X] Fix chromatic issues (Center team member fts and add the same size for expenditure columns)


YasielCabrera (approved):
lgtm

PR #1650, opened 2024-06-07
[Reskin-440] Add mobile padding for category and status

alejandrocabriales:
## Ticket

## What solved
- [X] Should change the chip's properties (categories, scopes) to small resolutions.
- [X] All categories should have the defined border. **Current Output:** Scope Facilitator is visible without border.


YasielCabrera (approved):
lgtm

PR #1661, opened 2024-06-10
[Fix-440] Fix the background in the filter when its open

alejandrocabriales:
## Ticket

## What solved
- [X] :Filters. **Expected Output:** The filters should be displayed as Ecosystem Actors shows. **Current Output:** By default the "All" option is displayed with dark background.


YasielCabrera (approved):
lgtm

PR #1664, opened 2024-06-11
Reskin/447 apply reskin design to the cu ea about

alejandrocabriales:
## What solved
- [X] CU/EA. Should update the "What we do" section for light/dark mode.Desktop/Small  resolutions.
- [X] Should update the Team Size section for light/dark mode. Desktop/Small  resolutions.


YasielCabrera (approved):
lgtm

PR #1845, opened 2024-08-06
Fusion/feat/hom 1.5 roadmap section

nesdc88:
## Ticket
https://trello.com/c/MFWdyfez/540-hom-15-roadmap-section-data

## Description
HOM-1.5 Homepage Roadmap Section

## What solved

- [X] Should add the coordinators

## Checklist:

- [X] I have performed a self-review of my own code
- [X] I have performed a self-review of my own chromatic changes
- [X] I have commented on my code, particularly in hard-to-understand areas
- [X] I have checked my code and corrected any misspellings
- [X] I have removed any unnecessary console messages
- [X] I have removed any commented code
- [X] I have checked that there are no buggy stories in Storybook

### DIFF
--- a/src/views/Home/components/MilestoneCard/MilestoneCard.tsx
+++ b/src/views/Home/components/MilestoneCard/MilestoneCard.tsx
@@ -5,8 +5,6 @@ import AvatarPlaceholderIcon from 'public/assets/svg/avatar_placeholder.svg';
 
 import Card from '@/components/Card/Card';
 import InternalLinkButton from '@/components/InternalLinkButton/InternalLinkButton';
-import SESTooltip from '@/components/SESTooltip/SESTooltip';
-import Information from '@/components/icons/information';
 
 import { siteRoutes } from '@/config/routes';
 import type { Maybe } from '@/core/models/interfaces/generics';
@@ -17,12 +15,11 @@ import { progressPercentage } from '@/views/RoadmapMilestones/utils';
 
 import useMilestoneCard from './useMilestoneCard';
 
-import type { FC, MutableRefObject } from 'react';
+import type { FC } from 'react';
 
 interface MilestoneCardProps {
   slug: string;
   milestoneData: Milestone;
-  coordinatorsRef: MutableRefObject<HTMLDivElement[]>;
 }
 
 interface ElementWithProgress {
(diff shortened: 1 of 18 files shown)


src/views/Home/components/MilestoneCard/MilestoneCard.tsx
```diff
@@ -69,6 +73,29 @@ const MilestoneCard: FC<MilestoneCardProps> = ({ slug, milestoneData }) => {
           <ProgressLabel progress={progress}>{usLocalizedNumber(progress * 100, 0)}%</ProgressLabel>
         </ProgressBarContainer>
       </Progress>
+      <CoordinatorsContainer
+        ref={(element) => {
+          if (element !== null) {
+            coordinatorsRef.current.push(element);
```
YasielCabrera, line 79: If we push the element here, we could be adding the element multiple times on every render

src/views/Home/components/RoadmapSection/useRoadmapSection.tsx
```diff
@@ -21,12 +21,24 @@ const useRoadmapSection = (roadmapsData: Roadmap[]) => {
     }
   };
 
+  const coordinatorsRef = useRef<HTMLDivElement[]>([]);
+
+  useLayoutEffect(() => {
```
YasielCabrera, line 26: For this case, do you see any issue with using `useEffect` instead? ` useLayoutEffect` tends to affect the performance and block the browser to paint the UI, so if we can avoid it, would be the best

YasielCabrera (changes requested):

src/views/Home/components/MilestoneCard/MilestoneCard.tsx
nesdc88: This definitely needs to be done very carefully... I've changed the approach to adjust sections heights to a simpler more "pure js" solution based on class names. This can be improved in the future...

src/views/Home/components/RoadmapSection/useRoadmapSection.tsx
nesdc88: Changed to useEffect :)

YasielCabrera (approved):
lgtm

